PR #5, opened 2015-03-15
Fix typo in config variable in README.

simonalpha:
Simple typo, but caused me a bit of frustration!


dannon:
Thanks for catching this!

PR #7, opened 2015-12-13
Correct default settings and extend settable options

pkrog:
Correct default paths to config directory.

Allow setting of options in section [server:main] of galaxy.ini, from YAML file.


natefoo:
Thanks @pierrickrogermele. A couple comments:

1. It is already possible to override `[server:main]` by specifying a `server:main` key in `galaxy_config`. The template simply has a "default" `[server:main]` if no other `server:` section is specified in `galaxy_config`.
2. I prefer to not ignore `*.swp` as this indicates a file is open by vim for writing. It can help to indicate that something is unfinished and should be addressed before committing.

PR #8, opened 2015-12-16
Dependencies are not installed if galaxy_fetch_eggs=False

fabiorjvieira:


natefoo:
I believe the logic here is correct. The default for both settings is `yes` (fetch). Disabling either one should prevent dependencies from being fetched, switching this to an `or` would not do that, you would have to explicitly disable both settings.

mvdbeek:
That's right, the only thing a little confusing is that `fetch_eggs` needs to be `yes`, even if you want to fetch wheels. But I think we can close this, since @fabiorjvieira is not working on this anymore.

PR #9, opened 2016-01-24
Fix pip at 8.0.0

mvdbeek:
Since pip seems to be evolving rapidly as of recent. 
Otherwise we're getting the [import errors](https://travis-ci.org/ARTbio/ansible-artimed/builds/104456943#L1663) on new instances.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,6 +13,9 @@ galaxy_manage_mutable_setup: yes
 galaxy_manage_database: yes
 galaxy_fetch_dependencies: yes
 
+# Set a version of pip to install
+galaxy_pip_version: 8.0.0
+
 # legacy
 galaxy_fetch_eggs: yes
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -14,7 +14,7 @@ galaxy_manage_database: yes
 galaxy_fetch_dependencies: yes
 
 # Set a version of pip to install
-galaxy_pip_version: 8.0.0
+galaxy_pip_version: 8.0.2
 
 # legacy
 galaxy_fetch_eggs: yes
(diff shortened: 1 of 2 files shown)


mvdbeek:
@natefoo maybe you have a better idea?

natefoo:
`--no-index --find-links https://wheels.galaxyproject.org/simple/pip` should prevent it from installing a newer pip from PyPI (unless that newer version is already installed). This was discussed on IRC today since that's exactly what the case will always be with Travis. The proposed solution is an "always install Galaxy pip even if it's a 'downgrade'" option, although I think we should probably make this the default as long as `--no-replace-pip` is not specified.

That said, I am not sure if it's possible with pip if you don't specify the exact version as part of `pip install` as you've done here, which I would prefer not to do because I want to be able to release bug fixes.

mvdbeek:
@natefoo yep, this was a workaround for travis ... I was first experimenting with downgrading the installed pip to 7.1.2 prior to running tests, but that didn't work. Perhaps the ansible pip module doesn't respect `--no-index` ?

natefoo:
Hrm, frustrating, and Ansible hides away the real failure. I believe that version should be `8.0.0+gx1`, but hold on a bit, I'm going to make a new `8.0.2+gx1` pip shortly.

mvdbeek:
Specifying `8.0.0` seems to always catch `8.0.0+gx1`, so at least this wouldn't prevent a potential `8.0.0+gx2` to get installed ...

natefoo:
Ah ok, interesting

natefoo:
@mvdbeek Can you update it to `8.0.2`? I think we're good to go then. You can also drop `--pre` from the pip args.

mvdbeek:
@natefoo done, thanks for looking into this!

natefoo:
Thank you for fixing it!

natefoo:
Just for the record, this won't be in galaxy.ansible.com until ansible/galaxy-issues#108 is resolved.

PR #13, opened 2016-06-08
Fix a few warnings

bgruening:


tasks/dependencies.yml
```diff
@@ -62,7 +62,7 @@
     extra_args: "--index-url https://wheels.galaxyproject.org/simple/"
     virtualenv: "{{ galaxy_venv_dir }}"
     virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
-  with_items: conditional_dependencies.stdout_lines
+  with_items: {{ conditional_dependencies.stdout_lines }}
```
mvdbeek: In the ansible docs items in with_* is always quoted:
http://docs.ansible.com/ansible/playbooks_loops.html
We should probably do that too.

bgruening:
@mvdbeek I have added quotes at least to the new ones. The rest can be part of a new PR I guess.

mvdbeek:
Thanks @bgruening

PR #14, opened 2016-06-15
Ensure tool_dependency_dir exists

mvdbeek:
This should make it easier to get started and I think this goes in the same direction as https://github.com/galaxyproject/galaxy/pull/2480.
Another place for this would be ansible-galaxy-tools, but I feel it makes most sense here.


mvdbeek:
anyone? @bgruening @jmchilton ?

mvdbeek:
thanks @jmchilton !

jmchilton:
Sorry for the delay @mvdbeek - I sort of think of this as Nate territory and didn't want to stomp on it - but I think he is hard at work making Galaxy packages.

mvdbeek:
No problem, I was under that impression too!

PR #29, opened 2018-07-09
galaxy_shed_tools_dir moved inside Galaxy clone

eslerm:
Moving to inside Galaxy clone conforms with line 33. Tested build.

Otherwise `tasks/static_setup.yml` may need to contain a new line 11 with `become: yes`.


eslerm:
It is odd to set `galaxy_shed_tools_dir` to be inside `galaxy_server_dir`, but I think it is a good default.

With this default, the Ansible role will not fail if the remote user does not have write permissions to the parent folder of `galaxy_server_dir`.

natefoo:
Thanks for the submission!

The goal with this path was to conform to the way that Galaxy itself was laid out with a standard `run.sh` start. I agree this path was bad - however, I've added a layouts feature in #49 that should fix this (but leaves the default as is in case anyone is using it).

PR #33, opened 2018-09-04
Updates for download.yml

eslerm:
Updated `galaxy_download_url` to correct path. e.g.,
  - From: `https://github.com/galaxyproject/galaxy.git/archive/release_18.05.tar.gz`
  - To:      `https://github.com/galaxyproject/galaxy/archive/release_18.05.tar.gz`

Added `makepyc.py` task which comes from `clone.yml`.

Replaced bash which downloads Galaxy and creates a receipt with Ansible modules.
  - In this version `galaxy_server_dir` is not removed. Directory removal was not added because, `clone.yml` does not remove the directory.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -22,7 +22,7 @@ galaxy_repo: "{{ galaxy_git_repo | default('https://github.com/galaxyproject/gal
 galaxy_commit_id: "{{ galaxy_changeset_id | default('master') }}"
 
 # Download URL (if used)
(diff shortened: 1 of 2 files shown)


natefoo:
Thanks for cleaning up and properly Ansiblizing this. With cloning, the git module will not update if there are local changes, unless you enable the force option (which will wipe out those changes). That said, any untracked or ignored files are not cleaned up. The addition of `makepyc.py` probably takes care of the biggest concern (`lib/`) so unless anyone voices any concerns, I can merge this.

PR #35, opened 2018-09-05
Added report for current and max database version

eslerm:
Reports the current and maximum database version before upgrading database.

Uses `clone.yml`'s report style.


mvdbeek:
Thanks @eslerm!

PR #36, opened 2018-09-14
change = to == and include_tasks to include for compatibility

drosofff:
with ansible==2.4 (retrocompatibily is ok with ansible 2.2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ the same play as this role:
       apt:
         name: "{{ item }}"
       become: yes
-      when: ansible_os_family = 'Debian'
+      when: ansible_os_family == 'Debian'
       with_items:
         - git
         - python-virtualenv
@@ -30,7 +30,7 @@ the same play as this role:
       yum:
         name: "{{ item }}"
       become: yes
-      when: ansible_os_family = 'RedHat'
+      when: ansible_os_family == 'RedHat'
       with_items:
         - mercurial
         - python-virtualenv
--- a/tasks/clone.yml
+++ b/tasks/clone.yml
@@ -32,7 +32,7 @@
     - email administrator with commit id
 
 - name: Include virtualenv setup tasks
-  include_tasks: virtualenv.yml
+  include: virtualenv.yml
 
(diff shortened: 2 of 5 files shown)


eslerm:
For the `ansible-galaxy-tools` repo I submitted a PR to change `include` to `include_tasks`. Ansible discourages include with error messages and in the documentation. Probably not backwards compatible though :(

> Include has some unintuitive behaviours depending on if it is running in a static or dynamic in play or in playbook context, in an effort to clarify behaviours we are moving to a new set modules (include_tasks, include_role, import_playbook, import_tasks) that have well established and clear behaviours. This module will still be supported for some time but we are looking at deprecating it in the near future.
[link](https://docs.ansible.com/ansible/2.6/modules/include_module.html)

Once extra nicety of `include_tasks` is that when a task is not included its sub-tasks do not show up in Ansible's standard output.

drosofff:
@eslerm you're right, I even saw after my PR that @natefoo moved the `include` to  `include_tasks` in this repo for the extra nicety you mentioned.

Yet, there is an issue I think. Sorry for the noise, I am going to investigate a bit more to see what I am missing.

By the way, this repo is not travis tested, don't you think it should be ?

natefoo:
> By the way, this repo is not travis tested, don't you think it should be ?

Yeah, it should be, I just haven't had the bandwidth to do so.

drosofff:
> Yeah, it should be, I just haven't had the bandwidth to do so.

I know what it is ! If I can give it a shoot I'll try ;-)

eslerm:
@drosofff, I've did some work on a travis build yesterday. This is my first time working on Travis. Would you mind helping me with it? I'm not sure how to right tests :D

If you're interested I'll submit a PR after my commute. The Travis build works under the travis user (non-root and non-galaxy user!). Tests are not written and conditionals are not yet supported and there is at least one task that has an ansible issue outside of travis.  https://github.com/eslerm/ansible-galaxy/tree/travis

drosofff:
@eslerm sure, thank you !

PR #38, opened 2018-09-18
Travis-CI

eslerm:
A basic Travis-CI implementation which tests `ansible-galaxy`'s build process.

There are a few things left to do such as test conditional tasks and to write tests of the running build.
 
Using `root` as the Ansible `remote_user` fails with Travis (https://github.com/travis-ci/travis-ci/issues/1091). Therefore, become was added to this PR. The Ansible `become` statement allows a sudo user who is not `root` and does not have to be `{{ galaxy_user }}` to run the playbook. Tasks were also added to fix file permissions where needed.

The `become` half of this PR adds a security benefit (`root` user access is not needed over SSH and the `galaxy_user` who owns the Galaxy files does not need sudo) and makes sure that resulting files are owned by the `galaxy_user` instead of the `remote_user`.

To do before merge:
- test that the server can be started
- check that all files are owned by `{{ galaxy_user }}`
  - done manually
  - redirecting standard output to error does not fail travis 👎 
- update `README.md` playbook example
  - In the advanced example `s/become_user/galaxy_user/g` & test
- test conditional tasks (e.g., download.yml)
  - semi-dependent on other PRs
- check that the DB creation python scripts can be initially ran with sudo instead of as the `galaxy_user`
  - After running this PR on a VM, `sh run.sh` works but there may be issues in the DB still..?

Some of the commits include more info.

`.travis.yml` follows [ansible-galaxy-extras](https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/.travis.yml) & [ansible-galaxy-tools](https://github.com/galaxyproject/ansible-galaxy-tools/blob/master/.travis.yml) as templates.

Thanks @drosofff and @eschen42!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
 
 install:
   # Install Ansible.
-  - pip install ansible
+  - pip install ansible==2.4
 
   # Add ansible.cfg to pick up roles path.
   - printf '[defaults]\nroles_path = ../' > ansible.cfg
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+---
+language: python
+python: "2.7"
+sudo: required
+
+services:
+  - docker
+
+before_install:
+  - docker --version
+  - docker info
+  - export GALAXY_HOME=/home/galaxy
+  - export GALAXY_TRAVIS_USER=galaxy
+  - export GALAXY_UID=1450
+  - export GALAXY_GID=1450
+  - sudo groupadd -r $GALAXY_TRAVIS_USER -g $GALAXY_GID
+  - sudo useradd -u $GALAXY_UID -r -g $GALAXY_TRAVIS_USER -d $GALAXY_HOME -p travis_testing -c "Galaxy user" $GALAXY_TRAVIS_USER
+  - sudo mkdir $GALAXY_HOME
+  - sudo chown -R $GALAXY_TRAVIS_USER:$GALAXY_TRAVIS_USER $GALAXY_HOME
+  - sudo chmod -R 775 $GALAXY_HOME
+
+install:
+  # Install Ansible.
+  - pip install ansible
+
+  # Add ansible.cfg to pick up roles path.
+  - printf '[defaults]\nroles_path = ../' > ansible.cfg
+
+script:
+  - ansible-playbook -i "localhost, " -c local tests/test_playbook.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
 
 install:
   # Install an upgraded pip and virtualenv
+  - sudo apt-get install git
   - sudo apt-get install python-pip
   - pip install virtualenv
   - pip install --upgrade pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - sudo useradd -u $GALAXY_UID -r -g $GALAXY_TRAVIS_USER -d $GALAXY_HOME -p travis_testing -c "Galaxy user" $GALAXY_TRAVIS_USER
   - sudo mkdir $GALAXY_HOME
   - sudo chown -R $GALAXY_TRAVIS_USER:$GALAXY_TRAVIS_USER $GALAXY_HOME
-  - sudo chmod -R 775 $GALAXY_HOME
+  - sudo chmod -R 755 $GALAXY_HOME
 
 install:
   # Install Ansible.
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,7 @@ language: python
 python: "2.7"
 sudo: required
 
-services:
-  - docker
-
 before_install:
-  - docker --version
-  - docker info
   - export GALAXY_HOME=/home/galaxy
   - export GALAXY_TRAVIS_USER=galaxy
   - export GALAXY_UID=1450
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,11 @@ before_install:
   - sudo chmod -R 755 $GALAXY_HOME
 
 install:
+  # Install an upgraded pip and virtualenv
+  - sudo apt-get install python-pip
+  - pip install virtualenv
+  - pip install --upgrade pip
+
   # Install Ansible.
   - pip install ansible==2.4
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,6 +13,11 @@ galaxy_manage_database: yes
 galaxy_fetch_dependencies: yes
 galaxy_manage_errordocs: no
 
+# Folder ownership and permission variables
+galaxy_user: galaxy
+galaxy_group: galaxy
+galaxy_mode: 0755
+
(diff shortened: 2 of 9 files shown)


martenson:
reopening immediately to trigger travis

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+before_install:
+  - export GALAXY_HOME=/home/galaxy
+  - export GALAXY_TRAVIS_USER=galaxy
+  - export GALAXY_UID=1450
+  - export GALAXY_GID=1450
+  - sudo groupadd -r $GALAXY_TRAVIS_USER -g $GALAXY_GID
+  - sudo useradd -u $GALAXY_UID -r -g $GALAXY_TRAVIS_USER -d $GALAXY_HOME -p travis_testing -c "Galaxy user" $GALAXY_TRAVIS_USER
+  - sudo mkdir $GALAXY_HOME
+  - sudo chown -R $GALAXY_TRAVIS_USER:$GALAXY_TRAVIS_USER $GALAXY_HOME
+  - sudo chmod -R 775 $GALAXY_HOME
```
drosofff: @eslerm shouldn't it be `755` ?

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+
+install:
+  # Install Ansible.
+  - pip install ansible
```
drosofff: I would pin ansible version
`pip install ansible==2.4`

.travis.yml
eslerm: yes

.travis.yml
eslerm: added

eslerm:
@drosofff reviews accepeted.

I will work on using the GitHub review process correctly.

drosofff:
@eslerm worked on a GCE VM and your test playbook

- [x] test that the server can be started
- [x] check that all files are owned by {{ galaxy_user }}
did it using `$ find . ! -user galaxy` and `$ find . ! -group galaxy`. Fine as soon as the first sh run.sh is performed by `galaxy` user (otherwise the files created at first Galaxy run are owned by the run user)
- [ ] redirecting standard output to error does not fail travis 👎
This I don't see what you mean ~

On another note: I think you should add as `pre-tasks` the install of python-virtualenv (and git even if it is most of times already present on target machines, it does not hurt) in your test playbook

drosofff:
@eslerm for testing galaxy, do you plan something like [this](https://github.com/ARTbio/GalaxyKickStart/blob/master/.travis.yml#L44), and subsequent huge battery of test ? If Yes, a Galaxy admin user has to be created (at least).

There is also the possibility to create a docker image with the ansible-galaxy role and use it for the testing (maybe overkilling, @natefoo ?)

Or just spin off Galaxy with the role and curl http://localhost:8080/api/version ?

eslerm:
GalaxyKickStart looks great! I might use it instead.

Is there a way to register an admin user and generate an API key with curl? Running ansible-galaxy-tools after ansible-galaxy requires that key.

If we add Docker, interoperability could be tested. I'd like to make sure CentOS works.

Should git and virtualenv installation be added to the role?

I was hoping that `echo "send stdio to stderr" 1>&2` would make Travi fail. Then if there was any output for your `find` command Travis would not pass. I should read their docs on error testing :)

drosofff:
> GalaxyKickStart looks great! I might use it instead.
Well thanks... but if we start to do that we are going to test also the role that GKS is using ;-)
 
> Is there a way to register an admin user and generate an API key with curl? Running ansible-galaxy-tools after ansible-galaxy requires that key.
It would be great if it is possible. I don't know I will try. @natefoo @mvdbeek any sights on this ?
 
> If we add Docker, interoperability could be tested. I'd like to make sure CentOS works.
If we could have generic docker images for testing all galaxyproject ansible roles, it would be nice. Maybe worth to think about it

> Should git and virtualenv installation be added to the role?
I don't want to be too "radical", but a test should be a test, not a "test in Travis"  ;-) What I said is that if you run your test playbook in a fresh ubuntu install, it will complain for sure about the lack of python-virtualenv. I did it and ansible failed to pass the task with the virtual environment installation.

> I was hoping that echo "send stdio to stderr" 1>&2 would make Travi fail. Then if there was any output for your find command Travis would not pass. I should read their docs on error testing :)
Oh no, I think the find command would be ok in Travis. Easy to test !

mvdbeek:
> Is there a way to register an admin user and generate an API key with curl? Running ansible-galaxy-tools after ansible-galaxy requires that key.

You can set the master_api_key and use that one. it's not a full user, but for installing tools that is sufficient (`galaxy_config_master_api_key`).

> Should git and virtualenv installation be added to the role?

Not to the role itself, we'd do this normally in ansible-galaxy-os

> I was hoping that echo "send stdio to stderr" 1>&2 would make Travi fail. Then if there was any output for your find command Travis would not pass. I should read their docs on error testing :)

Missing the context here, but if things in the script section return a non-zero exit code that will be a travis error, which should be sufficient, right ?

I really like the effort here, but `become` shouldn't be part of this role that doesn't otherwise require any privileged actions. What was the motivation for this ? If this is for travis testing it feels wrong to adjust functionality to the test. If there is no unprivileged user on travis can you not create one ?

eslerm:
@drosofff I'll start testing Ubuntu VMs also.

@mvdbeek thanks for the feedback. I want to make sure my PRs are useful to the Galaxy Devs.

Glad to know about the master api key!

I added become because I ran into issues with travis without it.

Originally I wanted to make a separate `travis` and `become` PR. My motivation for the standalone become branch was to allow an Ansible `remote_user` who is not `root` and does not have to be `{{ galaxy_user }}` to run the playbook for security. I'd rather have Ansible log into SSH with this sort of sudo user.

I went back and removed the become commits and ran a test with the `remote_user` set to `galaxy` in the playbook. [This failed with permission errors](https://travis-ci.org/eslerm/ansible-galaxy/builds/431176845).  I also tested [setting the user in the playbook](https://travis-ci.org/eslerm/ansible-galaxy/builds/431187352). Running with the playbook with [`travis` as the `remote_user` also fails](https://travis-ci.org/eslerm/ansible-galaxy/builds/431189344). [Adding `travis` to the `galaxy` group also fails](https://travis-ci.org/eslerm/ansible-galaxy/builds/431193775).

Is there a better solution I could try?

mvdbeek:
> Originally I wanted to make a separate `travis` and `become` PR. My motivation for the standalone become branch was to allow an Ansible `remote_user` who is not `root` and does not have to be `{{ galaxy_user }}` to run the playbook for security

You can set become at the play level, that solves the problem for the tests, and I imagine that should be fine for what you're trying as well, or oyu could specify the remote_user as is done in the usegalaxy playbook: https://github.com/galaxyproject/usegalaxy-playbook/blob/master/env/common/config.yml#L5

mvdbeek:
(I opened a PR with passing tests against this branch at https://github.com/eslerm/ansible-galaxy/pull/5)

eslerm:
Thank you!

mvdbeek:
Cool, with those last changes I'm fine merging this without @natefoo's review.

PR #39, opened 2018-11-05
Update print statement syntax to support Python 3.

ic4f:
Adding parentheses in 2 locations is enough to use role under Python 3.


bgruening:
Cool thanks @ic4f!

PR #41, opened 2018-11-08
Update deprecated syntax

ic4f:
Invoking "apt" only once while using a loop via squash_actions is
 deprecated. Use list of items as name value instead.


bgruening:
Thanks @ic4f!

PR #43, opened 2018-11-10
[WIP] Add custom filter to solve duplicate key issue (issue 42)

ic4f:
See issue 42 for problem description + comments in seqtodict.py
Simplest way to run the unit tests:  `python -m unittest discover` from root dir.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,4 @@ install:
 
 script:
   - ansible-playbook -i "localhost, " -c local tests/test_playbook.yml
+  - python -m unittest tests.test_seqtodict
--- a/README.md
+++ b/README.md
@@ -198,8 +198,9 @@ and start two [job handler mules][deployment-options].
           - unix_signal:15 gracefully_kill_them_all
         py-call-osafterfork: true
         enable-threads: true
-        mule: lib/galaxy/main.py
-        mule: lib/galaxy/main.py
+        mules:
+          - mule: lib/galaxy/main.py
+          - mule: lib/galaxy/main.py
         farm: job-handlers:1,2
       galaxy:
         database_connection: "postgresql:///galaxy?host=/var/run/postgresql"
--- a/filter_plugins/seqtodict.py
+++ b/filter_plugins/seqtodict.py
@@ -1,0 +1,94 @@
+import re
+
+from ansible.errors import AnsibleFilterError
+
+def get1():
+    return 1
+
+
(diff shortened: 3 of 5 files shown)


ic4f:
Sorry for the failing tests in prev.commits - I hadn't setup travis for those, so I missed them (that's what happens when committing at 3am...) All fixed now. 
@natefoo : please take a look: this should address a minor problem resulting from commit b69cc10 (Jun 22), lines 201-202).

ic4f:
So, apparently, uwsgi_yaml takes care of this already. The 2 filters are different: uwsgi_yaml takes in a dictionary and "flattens" ALL sequences, whereas seqtodict takes in a formatted yaml string and 'flattens" only those sequences that have a given KEY. Still, given that we have uwsgi_yaml, seqtodict is not needed at this time,  so I'm closing this PR.

There's still an error in the sample code (see advanced playbook: duplicate keys/mule). I'll submit a separate PR for that.

eslerm:
> There's still an error in the sample code (see advanced playbook: duplicate keys/mule). I'll submit a separate PR for that.

Each mule creates a mule worker in uWSGI syntax https://uwsgi-docs.readthedocs.io/en/latest/Mules.html

ic4f:
> > There's still an error in the sample code (see advanced playbook: duplicate keys/mule). I'll submit a separate PR for that.
> 
> Each mule creates a mule worker in uWSGI syntax https://uwsgi-docs.readthedocs.io/en/latest/Mules.html
Yes, which make it a valid config for uwsgi. However, Ansible expects legal YAML, which does not allow duplicate dictionary keys. As a result, Ansible with process *only the last key*. (you can see that message when you run that playbook - and I hope I'm not missing anything here..)

eslerm:
That sounds right.

There is a variable for using real yaml if you have a libyaml compiled version of uWSGI. That doesn't fix the readme though :)

ic4f:
> That sounds right.
> 
> There is a variable for using real yaml if you have a libyaml compiled version of uWSGI. That doesn't fix the readme though :)

Yes, it's the ```galaxy_uwsgi_yaml_parser``` option. However, the playbook itself is processed by ansible (which uses libyaml), so it has to be valid yaml. So the parser option sets the format for the *output*  configuration for uwsgi, but has no effect on how the input playbook code is parsed by ansible.
And thanks for looking at this - I'm not at all sure I'm not missing stuff, so an extra pair of eyes is very helpful!

PR #45, opened 2018-11-23
Make checking repo dir safe for --check

erasche:
Using commands meant it wouldn't run during `ansible-playbook --check` which was troublesome for us.


natefoo:
Would this be fixed by #53?

erasche:
Yes that will work

PR #47, opened 2019-01-21
Changes from EU

erasche:
- Permit running with `--check` mode by replacing commands with git module, and disabling unnecessary tasks
- Permit disabling backups
- Sort the config file before writing.
- Common startup on git version change.


README.md
```diff
@@ -267,8 +267,9 @@ Author Information
 
 This role was written and contributed to by the following people:
 
-[Enis Afgan](https://github.com/afgane)  
-[Dannon Baker](https://github.com/dannon)  
-[Simon Belluzzo](https://github.com/simonalpha)  
-[John Chilton](https://github.com/jmchilton)  
-[Nate Coraor](https://github.com/natefoo)  
+[Enis Afgan](https://github.com/afgane)
```
natefoo: Trivial request, sorry. Can you retain the two spaces at the end of each line? Without those spaces in markdown these collapse to a single line.

tasks/build_client.yml
```diff
@@ -0,0 +1,14 @@
+---
+# First ensure that nodejs is installed into the venv by running common_startup.
+- name: Execute common_startup.sh
```
natefoo: Thus far I've tried to recreate `common_startup.sh` in the role rather than call it directly since it gives the role user more control and is more Ansible-ic. I can do this for the client build if you want to focus elsewhere (I don't want to waste your time on my inanity). Bonus: [tasks are already written](https://github.com/galaxyproject/usegalaxy-playbook/blob/master/roles/usegalaxy_cvmfs/tasks/_inc_client_build.yml).

README.md
erasche: Trivial requests are always welcome :) Would it be ok to just make it a proper markdown list?

natefoo: Alternatively these could be a list (since they are, like, a list of things).

tasks/build_client.yml
natefoo: #52

erasche: Next time could you just add to my branch? Or PR to my branch if you don't want to add directly? That way I don't have to go back and re-edit this PR to make it merge conflict safe?

PR #48, opened 2019-01-23
setting depth param for clone task + minor old PR

ic4f:
Setting the depth param (=1) creates a shallow clone reducing the size of the .git/ directory from 467M to 22M. We need this for constructing a minimal galaxy container image  (see issue #7225).

The parameter is omitted when not defined. 

If this change is OK, I'll add a note about it to the role's README.

### DIFF
--- a/tasks/clone.yml
+++ b/tasks/clone.yml
@@ -27,6 +27,7 @@
     repo: "{{ galaxy_repo }}"
     version: "{{ galaxy_commit_id }}"
     executable: "{{ git_executable | default(omit) }}"
+    depth: "{{ galaxy_repo_clone_depth | default(omit) }}"
   notify:
     - restart galaxy
     - email administrator with commit id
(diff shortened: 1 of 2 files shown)


bgruening:
@ic4f if you do not need the git checkout you could consider using simply wget and be even smaller. See here: https://github.com/bgruening/docker-galaxy-stable/blob/master/compose/galaxy-init/Dockerfile#L34

erasche:
ansible-galaxy supports a `wget` with the [`download` option](https://github.com/galaxyproject/ansible-galaxy/blob/master/tasks/main.yml#L13) instead of the `git` clone.

ic4f:
Thank you, @bgruening, @erasche . Yes, it seems that's all we need, really. And the comments for the initial commit of the download tasks file by @jmchilton actually describes our use case: 
> ...a way to get Galaxy without the whole history for throw away machines like Docker - where keeping the size of the image small is more important than being able to update the Galaxy found inside. (7ed9bbf)
Thank you, all. Closing the PR.

Issue #42, opened 2018-11-10
Duplicate dictionary keys not allowed


ic4f:
Accoring to this project's [README](https://github.com/galaxyproject/ansible-galaxy/blob/master/README.md) and [Galaxy documentation](https://docs.galaxyproject.org/en/master/admin/scaling.html#deployment-options), the following is part of a valid playbook configuration:
```yml
uwsgi:
  mule: lib/galaxy/main.py
  mule: lib/galaxy/main.py
```

However, this does not work: Ansible expects legal YAML, which does not allow duplicate dictionary keys. As a result, **only the last key will be processed**.

This would work if these were items in a sequence:
```yml
uwsgi:
  some-key:
    - mule: lib/galaxy/main.py
    - mule: lib/galaxy/main.py
```

However, that changes the format of the output configuration used by uwsgi.

One solution is to transform Ansible's output, flattening the sequence by "pulling it" up one level and removing its key. This can be done with a custom Ansible plugin. I'm submitting a pull request with a possible solution (this could be overkill if the uwsgi format can be easily changed to expect a sequence of mules, or other duplicate keys... )

ic4f:
The underlying issue is taken care of by the uwsgi_yaml filter (see comments on closed PR #43 )

There's still an error in the sample code in README.md (see advanced playbook: duplicate keys/mule). I'll submit a separate PR for that.

I'm closing this issue.

PR #50, opened 2019-01-25
fix mule in advanced example in README.md

Banshee1221:
As referenced by @ic4f in issue #42, the example stating:

```yaml
mules:
  - mule: lib/galaxy/main.py
  - mule: lib/galaxy/main.py
```
Is incorrect and should rather be:
```yaml
mule:
  - lib/galaxy/main.py
  - lib/galaxy/main.py
```

This appears to work correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,9 +197,9 @@ and start two [job handler mules][deployment-options].
           - unix_signal:15 gracefully_kill_them_all
         py-call-osafterfork: true
         enable-threads: true
-        mules:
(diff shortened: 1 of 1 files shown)


natefoo:
Thanks!

ic4f:
Thanks! 
(That syntax was contingent on a different filter (as per issue #43 ) that relied on that format to convert to uwsgi yaml, but ended up not being used.)

PR #54, opened 2019-01-26
Added pip_extra_args parameter

nuwang:
Added extra args parameter to allow passing custom parameters such as --compile and --no-cache-dir to pip.


nuwang:
@natefoo This was done to support: https://github.com/CloudVE/galaxy-kube-playbook/pull/2 but I'm not sure whether the --compile flag is necessary to generate .pyc files - doesn't look like it. Would you also be able to comment on that PR?

PR #55, opened 2019-01-27
Fixes for privilege separation

natefoo:


natefoo:
For future @natefoo, to expand a bit on the comment I left in there:

The reason the `default(omit)`s were changed was because using them caused the existing value of `become_user` as set on the role, play, etc. to be *lost* - `default(omit)` essentially *removed* the `become_user` param if the default condition occurred.

The reason `is not undefined` was removed is more because that implementation was a mistake on my part, it didn't even attempt to preserve the existing value of `become`. However, the replacement (looking at `ansible_env.SUDO_USER`) is an enormous hack that is necessary because, astonishingly, Ansible provides **no** mechanisms for determining the value of `become`. Not a fact, not a var, *nothing*.

Finally, `ansible_user_id` is essentially the value obtained by Ansible running `id -un` which works for our purposes, but there is *also* no way to know what user Ansible connected as, if it has since used `become` except by inspecting `$SUDO_*` env vars. And it turns out that `ansible_user` can be undefined, it's only set when you set `ansible_user` in the inventory, or `remote_user` on the play, role, task, etc. If you just let Ansible connect as whatever user `ssh` uses by default (either your username or anything you've configured in `~.ssh/config`), `ansible_user` is unset. So it's pretty much useless.

PR #60, opened 2019-02-07
(1) Fix typo in sample code; (2) Selection of client build type

ic4f:


defaults/main.yml
```diff
@@ -249,3 +249,11 @@ galaxy_uwsgi_config_default:
 
 # TODO: Implement something like this for starter/restarter
 #galaxy_server_type: paste_singleprocess
+
+#
+# Client build settings
+#
+
+# Options include: client / client-production (default) / client-production-maps  
+galaxy_client_make_target: client-production
```
erasche:
```suggestion
galaxy_client_make_target: client-production-maps
```

maybe we should continue to use the same value that was previously used, to not cause unexpected changes for people using this role who had not set that variable and were expecting the default to be that.

ic4f: I thought none of these values were previously used: the build client option was added to the role quite recently (#52), so there would be no unexpected changes for people using this role. The values are consistent with the client readme file. Sorry, I hope I'm not missing something here..

erasche: We updated the role for use in an admin training we gave, everyone using these roles has opted into that behaviour, so unfortunately potentially people are relying on that.

I imagine @natefoo chose that default for a reason, but maybe he has an opinion on this?

ic4f: OK, so you mean make client-production-**maps** as the default? I have no objection, as long as client-production and client are available as options.

erasche: Yeah, exactly, just the value that was previously hardcoded. Of course everything should stay available

ic4f: Got it. Will change in a minute.

ic4f: Thank you, @erasche!

erasche:
Thanks @ic4f!

defaults/main.yml
natefoo: For the record, I picked that target because, iirc, that's what @martenson told me should be used for production.

martenson: I still think our servers should build and serve js maps so `client-production-maps` is the proper target.

PR #62, opened 2019-02-21
Replace: pip_virtualenv_command > galaxy_virtualenv_command

ic4f:
Keep backward compatible.

### DIFF
--- a/tasks/dependencies.yml
+++ b/tasks/dependencies.yml
@@ -13,7 +13,7 @@
         requirements: "{{ galaxy_requirements_file }}"
         extra_args: "--index-url https://wheels.galaxyproject.org/simple/ --extra-index-url https://pypi.python.org/simple {{ pip_extra_args | default('') }}"
         virtualenv: "{{ galaxy_venv_dir }}"
-        virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
+        virtualenv_command: "{{ galaxy_virtualenv_command | default(pip_virtualenv_command | default(omit)) }}"
       environment:
         PYTHONPATH: null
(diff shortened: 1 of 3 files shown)


tasks/dependencies.yml
```diff
@@ -13,7 +13,7 @@
         requirements: "{{ galaxy_requirements_file }}"
         extra_args: "--index-url https://wheels.galaxyproject.org/simple/ --extra-index-url https://pypi.python.org/simple {{ pip_extra_args | default('') }}"
         virtualenv: "{{ galaxy_venv_dir }}"
-        virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
+        virtualenv_command: "{{ galaxy_virtualenv_command | default(pip_virtualenv_command | default(omit)) }}"
```
drosofff: I think this change should be accompanied by setting the pip_virtualenv_command to something in default/main.yml  for backward compatibility ('virtualenv' for instance). But I do not really understand the purpose of `default(omit)`

mvdbeek: The [omit filter](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#omitting-parameters) skips module parameters. The module parameter is `virtualenv_command`, and the module is pip. So this is equivalent to not specifying `virtualenv_command`. This defaults to `virtualenv` if not specified (https://docs.ansible.com/ansible/latest/modules/pip_module.html). So in this case you can either use `galaxy_virtualenv_command ` (preferred) or `pip_virtualenv_command ` (fallback) or nothing.

drosofff:
I see... I am testing the theory ;-) but as Travis is not stable these days I'll wait before drawing conclusions. I'll keep you posted !

PR #63, opened 2019-03-05
Add spinning fifo plugin

erasche:
Completely untested.


erasche:
Unused, uninteresting.

PR #64, opened 2019-04-05
Potential simplification for dir permissions

erasche:
When trying to do priv sep, it's restricted enough that nginx cannot access static files. While adding nginx to the galaxy group is one option, it can be a bit messy in ansible? Maybe this just makes things easier?


erasche:
Fixed this by override permission variable in extra vars. Neat.

PR #67, opened 2019-04-15
Fix client generation when galaxy was installed from archive

abretaud:
Hi,
Here's a little patch to fix a problem found while updating https://github.com/galaxyproject/planemo-machine: as galaxy is installed from a downloaded archive, the client generation failed.
My artistic level in ansible is not very high, so let me know if there's a better way to do it!
I will make a PR for planemo-machine soon


jmchilton:
Thanks for fixing this, awesome!

PR #80, opened 2019-07-05
Template out shed tool conf more better

erasche:
xref https://github.com/galaxyproject/training-material/pull/1445

does this look ok? untested because internet.


natefoo:
I think this would overwrite shed_tool_conf.xml after Galaxy writes to it?

erasche:
@natefoo addressed with `force: no`

PR #85, opened 2019-09-27
Deconstruct client build process

natefoo:
This adds some transparency for the admin but requires us to synchronize with the client build scripts and will be complicated by differences across versions. Because of that, the default is still to invoke `make client-production-maps`.

### DIFF
--- a/tasks/_inc_client_build_steps.yml
+++ b/tasks/_inc_client_build_steps.yml
@@ -9,16 +9,20 @@
     VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
 
 - name: Run gulp
-  command: gulp
+  command: yarn run gulp {{ item }}
   args:
     chdir: "{{ galaxy_server_dir }}/client"
+  with_items:
+    - fonts
+    - stageLibs
+    - plugins
   environment:
     PATH: "{{ galaxy_server_dir }}/client/node_modules/.bin:{{ galaxy_venv_dir }}/bin:{{ ansible_env.PATH }}"
     VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
     NODE_ENV: production
 
 - name: Run webpack
-  command: webpack -p
+  command: yarn run webpack-production
   args:
     chdir: "{{ galaxy_server_dir }}/client"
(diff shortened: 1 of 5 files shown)


tasks/_inc_client_build_steps.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Install packages with yarn
+  yarn:
+    executable: "yarn --network-timeout 300000 --check-files"
+    path: "{{ galaxy_server_dir }}/client"
+  environment:
+    PATH: "{{ galaxy_venv_dir }}/bin:{{ ansible_env.PATH }}"
+    VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+
+- name: Run gulp
+  command: gulp
```
dannon: This relies on the gulp cli, right?  I'd probably use the package scripts directly and rely on `yarn run gulp` here.

tasks/_inc_client_build_steps.yml
```diff
@@ -0,0 +1,32 @@
...
+  args:
+    chdir: "{{ galaxy_server_dir }}/client"
+  environment:
+    PATH: "{{ galaxy_server_dir }}/client/node_modules/.bin:{{ galaxy_venv_dir }}/bin:{{ ansible_env.PATH }}"
+    VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+    NODE_ENV: production
+
+- name: Run webpack
+  command: webpack -p
```
dannon: Same as above; `yarn run webpack-production` which will apply appropriate flags (right now just -p, but who knows down the road)

dannon:
If you did want to separate the (new, and likely growing) plugin building/staging from the other gulp stuff that's central to the actual client build you could break it down like this:

`yarn run gulp fonts && yarn run gulp stageLibs`
`yarn run gulp plugins`

Default does all of that, and for simplicity the 'plugins' one is lumped into the same process, but I wasn't sure if you'd want to separate those concerns a little more in this decomposition.  Doesn't matter to me.

natefoo:
Thanks for the review and suggestions!

PR #90, opened 2019-10-25
Install libyaml version of uwsgi when galaxy_uwsgi_yaml_parser == libyaml

nuwang:
**Background**
The original issue is detailed here: https://github.com/galaxyproject/galaxy/issues/8558

The TL; DR; version is that uwsgi has two config parsers, an internal parser (which uses an invalid yaml format) and a libyaml based parser (valid yaml but requires manual compilation with libyaml because the wheel version defaults to the internal parser).

**This PR**
Since we needed a valid yaml format for the k8s docker container, the original plan that was hatched in that issue was to make the required changes in the k8s Dockerfile here: https://github.com/CloudVE/galaxy-docker-k8s/pull/11. However, @almahmoud  later pointed out that  there was in fact a setting specifying that libyaml should be used for configs: `galaxy_uwsgi_yaml_parser`. If that setting is used for configs, we should clearly install the compatible libyaml, so we then thought that it's better to make the modification in ansible-galaxy itself.

We thought we'd found a nice and compact way to implement the solution: https://github.com/galaxyproject/ansible-galaxy/commit/b632350b4896d0283964036ed7baae853c757589, except for the minor detail that it didn't work because the galaxy index server only has the pyuwsgi wheel, and I don't know how you can get a source-compilable version to sit alongside it.

So unless @natefoo has some magic he could work here, the current PR contains a slightly lengthier workaround: https://github.com/galaxyproject/ansible-galaxy/commit/06fa8585ec69d7ff7a53beb7e02f748bda677570

### DIFF
--- a/files/libyaml.ini
+++ b/files/libyaml.ini
@@ -1,0 +1,27 @@
+# Settings for compiling libyaml version of UWSGI
+[uwsgi]
+main_plugin = python,gevent
+xml = auto
+yaml = libyaml
+json = auto
+ssl = auto
+pcre = auto
+routing = auto
+debug = false
+unbit = false
+malloc_implementation = libc
+extras =
+plugins =
+bin_name = uwsgi
+append_version =
+plugin_dir = .
+embedded_plugins = %(main_plugin)s, ping, cache, nagios, rrdtool, carbon, rpc, corerouter, fastrouter, http, ugreen, signal, syslog, rsyslog, logsocket, router_uwsgi, router_redirect, router_basicauth, zergpool, redislog, mongodblog, router_rewrite, router_http, logfile, router_cache, rawrouter, router_static, sslrouter, spooler, cheaper_busyness, symcall, transformation_tofile, transformation_gzip, transformation_chunked, transformation_offload, router_memcached, router_redis, router_hash, router_expires, router_metrics, transformation_template, stats_pusher_socket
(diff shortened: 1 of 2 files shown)


natefoo:
Are pyuwsgi and uwsgi both happily existing in the same venv? The uwsgi install is going to overwrite `bin/uwsgi` and future pyuwsgi upgrades in `requirements.txt` will break uwsgi. But I guess as long as you always install with the role it will always overwrite pyuwsgi with uwsgi?

nuwang:
Yes, seems to be working fine. However, valid point that if you do a manual upgrade without the role, the user has to watch out for the version. I was wondering, is it an option to change Galaxy's dependency from pyuwsgi to uwsgi in such a way that just doing a `pip install uwsgi` will do the right thing depending on whether the `UWSGI_PROFILE` is set or not? (while still using the Galaxy wheels server)

nuwang:
Closing this PR because ultimately, even the libyaml version of uwsgi handles valid yaml poorly. Instead, it seems better to just move the uwsgi config to a separate file and treat it as non-yaml, with the rest of the galaxy config being in galaxy.yml as usual. Example usage:
```
uwsgi --yaml /galaxy/server/config/uwsgi.yml --set galaxy_config_file=/galaxy/server/config/galaxy.yml
```

PR #94, opened 2019-10-28
Support deconstructed client build on > 19.09

### LINKED ISSUE #88: Make deconstructed client build conditional on version
Just as we now do for the node version, the commands to run for the deconstructed client build process should be conditional on Galaxy major version.

xref #85 galaxyproject/galaxy#8742

natefoo:
@dannon does this look right?

Fixes #88


dannon:
Yep!

natefoo:
Thanks!

PR #99, opened 2020-01-09
Add missing conf file templates

drosofff:
`shed_data_manager_conf.xml` and `shed_tool_data_table.xml` files were added in the template dir
and referenced in the ansible variable `galaxy_mutable_config_templates`.

Otherwise the server does not start using GalaxyKickStart. I suppose this would be the same with other means for deploying galaxy using this ansible role

### DIFF
--- a/tasks/virtualenv.yml
+++ b/tasks/virtualenv.yml
@@ -34,3 +34,4 @@
   environment:
     PYTHONPATH: null
     VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+  when: ansible_python_version is version('3', '<')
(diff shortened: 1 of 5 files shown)


drosofff:
Hi all, I think this PR is ready for review. I do not have access to the reviewers request, though

tasks/virtualenv.yml
```diff
@@ -24,3 +24,13 @@
   environment:
     PYTHONPATH: null
     VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+
+- name: pin setuptools to 44.0.0
+  pip:
+    name: setuptools
+    version: 44.0.0
+    virtualenv: "{{ galaxy_venv_dir }}"
+    virtualenv_command: "{{ galaxy_virtualenv_command | default(pip_virtualenv_command | default(omit)) }}"
```
mvdbeek: I would only do this on python 2.7, and I think this might be overridden by any of Galaxy's dependencies.

PR #101, opened 2020-01-28
Fix handling of files that are not longer present in Galaxy 19.09

pvanheus:
Galaxy 19.09 has changed how it installs "default" files, breaking the Ansible install. This PR aims to fix that.


pvanheus:
I've discovered a bug created by this PR and am working on a fix.

PR #103, opened 2020-02-10
Second attempt at solving post-19.09 mutable file setup

pvanheus:
This addresses problems not known when PR #101 was submitted. In short, in that PR 3 config files (migrated_tools_conf.xml, shed_tool_data_table_conf.xml and shed_data_manager_conf.xml) were ignored because they are not present in Galaxy 19.09. In addition, the shed_tool_conf.xml file was removed from the list of Galaxy config files. The result was a Galaxy that boots but when a new tool is installed it tries to write to a config directory that, in some directory layouts specified in this role, is read-only.

This PR takes a different approach and provides files or templates for the files that are no longer in 19.09 (as per suggestion from @mvdbeek) and ensures that files that will be changed by Galaxy are mentioned in the config and thus located in the mutable config directory. Tool installation thus succeeds.


defaults/main.yml
```diff
@@ -199,31 +202,39 @@ galaxy_app_config_default:
...
+    dest: "{{ galaxy_config_merged[galaxy_app_config_section].shed_tool_data_table_config }}"
+    
+# required for Galaxy release 19.09 and later
+galaxy_extra_mutable_config_templates:
+  - src: "migrated_tools_conf.xml.j2"
+    dest: "{{ galaxy_config_merged[galaxy_app_config_section].migrated_tools_config }}"
+
+# produce the galaxy_mutable_config_templates differently depending on Galaxy release
+# older versions used the value from galaxy_base_mutable_config_templates, newer versions combine that with galaxy_extra_mutable_config_templates
+galaxy_mutable_config_templates: "{{ (galaxy_base_mutable_config_templates + galaxy_extra_mutable_config_templates) if (galaxy_commit_id | replace('release_', '') | float() >= 19.09) else galaxy_base_mutable_config_templates | default([]) }}"
```
natefoo: This isn't guaranteed to work since the commit ID is not necessarily a release branch (or a branch at all). We have a task-based method for getting the major version in [tasks/_inc_galaxy_version.yml](https://github.com/galaxyproject/ansible-galaxy/blob/master/tasks/_inc_galaxy_version.yml). If you include this before you need `galaxy_mutable_config_templates` (but after clone/update tasks) you can use `__galaxy_major_version is version('19.09', '>=')` here.

natefoo:
It turns out this statement is not correct:
> I think we could get away with not instantiating these in the role if we had the role set defaults for these configs to paths in the mutable config dir in `galaxy_config_default`.

The *mutable config dir* was intentionally made to be not configurable in 19.09 because mutable configs are "Galaxy data," not things that an admin is supposed to directly modify. As such the idea was for those to live under the *data dir* in an (e.g. pip) installed Galaxy. When "running from source" (as all your traditional Galaxies will do) the mutable config dir is instead set to the config dir (i.e. `config/`). Overriding individual mutable config paths such as `shed_tool_config_file` is possible, but it also disables the automatic creation of that file - for good reason: if an admin is explicitly setting the path to that config, it is assumed that the file is expected to already exist, and continuing without it means already-installed tools would fail to load and the install DB could become out of sync with the shed tool config.

This all causes a problem for the privilege-separated role, since there's no way to both

1. separate the mutable (galaxy writable) configs from the static (galaxy read-only) configs, and
1. allow Galaxy to automatically generate mutable configs.

After some discussion on gitter, I believe the plan is to make the mutable config dir path configurable in 20.05, which means we'll probably want to include these changes in the role for 19.09 and 20.01.

PR #104, opened 2020-02-12
Default to omitting the virtualenv command, and don't force it to follow the format `python...`

natefoo:
This would impact galaxyproject/training-material#1659.

cc @hexylena


hexylena:
Looks OK, `pip_virtualenv_command` is the one that would affect a bunch of EU roles since we consume that variable if it's set.

PR #109, opened 2020-03-05
Rename mutable to managed

ic4f:
To stay consistent with Galaxy's terminology

Note: "manage 'managed'" is rather awkward and "manage 'galaxy-managed'" may be less so; but in the big picture 'galaxy-managed' creates more inconsistency even across this repo alone, so it's not worth it, IMHO.

Ref: https://github.com/galaxyproject/galaxy/pull/9444


ic4f:
Also, ref https://github.com/galaxyproject/galaxy/issues/9362

defaults/main.yml
```diff
@@ -12,7 +12,7 @@ galaxy_manage_paths: no
 galaxy_manage_clone: yes
 galaxy_manage_download: no
 galaxy_manage_static_setup: yes
-galaxy_manage_mutable_setup: yes
+galaxy_manage_managed_setup: yes
```
hexylena:
```suggestion
galaxy_manage_managed_setup: "{{ galaxy_manage_mutable_setup | default(yes) }}"
```

maybe something like this to ensure that people who don't read the upgrade notice can use the legacy value? @natefoo what are your opinions on that. "Read the release notes or else?"

ic4f: I have no objections. @natefoo - up to you?

natefoo: Yes, and we'll need to default all the other changed vars as well.

hexylena: Coolie. Sorry for the extra work, but figure it'll be a bit more userfriendly.

natefoo: It shouldn't be too bad as you can do:

```yaml
galaxy_managed_data_dir: "{{ galaxy_mutable_data_dir }}"
```

Rather than defaulting each instance of its usage.

PR #117, opened 2020-11-10
Added support for using an existing Galaxy folder

nuwang:
Adds a new parameter, `galaxy_manage_existing`, which allows an existing galaxy directory to be reused, instead of cloning from git or downloading an archive. In this case, `galaxy_server_dir` is considered to be the existing directory. This is useful for building Docker images from the current folder.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -7,8 +7,9 @@
 - name: Ensure that mutually exclusive options are not set
   assert:
     that:
-      - not (galaxy_manage_clone and galaxy_manage_download)
-    fail_msg: "Variables galaxy_manage_clone and galaxy_manage_download cannot both be true."
+      - [galaxy_manage_clone | bool, galaxy_manage_download | bool, galaxy_manage_existing | bool].count(True) == 1
+    fail_msg: |
+      "Only one of variables galaxy_manage_clone, galaxy_manage_download and galaxy_manage_existing can be true."
 
 # To my knowledge there is no other way to know whether or not the consumer of this role has set `become: true` outside
 # this role, and `become: {{ foo | default(omit) }}` on a task/block actually clobbers the higher level use of `become`.
@@ -44,6 +45,11 @@
   include_tasks: download.yml
   when: galaxy_manage_download
   tags: galaxy_manage_download
+
+- name: Include manage existing galaxy tasks
+  include_tasks: existing.yml
+  when: galaxy_manage_existing
+  tags: galaxy_manage_existing
 
 - name: Include static config setup tasks
   include_tasks: static_setup.yml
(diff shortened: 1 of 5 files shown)


tasks/main.yml
```diff
@@ -7,8 +7,9 @@
 - name: Ensure that mutually exclusive options are not set
   assert:
     that:
-      - not (galaxy_manage_clone and galaxy_manage_download)
-    fail_msg: "Variables galaxy_manage_clone and galaxy_manage_download cannot both be true."
+      - [galaxy_manage_clone | bool, galaxy_manage_download | bool, galaxy_manage_existing | bool].count(True) == 1
```
nuwang, line 10: @natefoo This syntax works ok in newer ansible versions, but it looks like it doesn't on ansible<2.8 on travis. Any thoughts on an alternative?

almahmoud: Not the most elegant, and I might have missed something cause my logic is rusty, but how about:
```
(not (galaxy_manage_clone and (galaxy_manage_download or galaxy_manage_existing))) and (not (galaxy_manage_download and (galaxy_manage_clone or galaxy_manage_existing))) 
```
I think that should cover all possible pairs?

hexylena: I'm fine with excluding <2.8, we keep these pretty updated, people should be updating their ansible, and this feels cleaner and easier to keep updated.

hexylena (approved):
Strange use case but it's opt in, so, doesn't hurt to merge this. Sorry it's been sitting around so long.

hexylena:
@nuwang could you please apply this patch? I think because you've opened the PR from your master branch I can't push back to it. In the future please open PRs from branches, opening from your master branch is an anti-pattern.

```
commit 343a54aa2aafa9e58e65d3663b3681684d298263 (HEAD -> nuwang-master)
Author: Helena Rasche <hxr@hx42.org>
Date:   Wed Jan 13 10:34:46 2021 +0100

    deprecate

diff --git a/.travis.yml b/.travis.yml
index abdb4e7..9611dca 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python

 matrix:
   include:
-  - python: "2.7"
   - python: "3.6"

 sudo: required
@@ -13,7 +12,7 @@ install:
   - sudo apt-get install git

   # Install Ansible
-  - pip install 'ansible<2.8'
+  - pip install 'ansible<2.10'

   # Add ansible.cfg to pick up roles path
   - printf '[defaults]\nroles_path = ../' > ansible.cfg
```

nuwang:
@hexylena My mistake, have made a new PR including your patch and have additional factored out the compile step based on @natefoo's suggestion.

nuwang:
Closing in favour of: https://github.com/galaxyproject/ansible-galaxy/pull/119

hexylena:
Thanks @nuwang!

PR #121, opened 2021-01-19
updated README according experience of installation

Areso:
1) without pointing to the galaxy prefix, it doesn't work
2) some bug (or feature), related to Ansible 2.10.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,9 +320,21 @@ Install Galaxy on your local system with all the default options:
     galaxy_server_dir: /srv/galaxy
   connection: local
   roles:
-     - galaxy
-```
-
+     - galaxyproject.galaxy
+```  
+  
+If your Ansible version >= 2.10.4, then  
+```yaml
+- hosts: localhost
+  vars:
+    galaxy_server_dir: /srv/galaxy
+    ansible_user: 'anyname'
+  connection: local
+  roles:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -320,9 +320,21 @@ Install Galaxy on your local system with all the default options:
...
+     - galaxyproject.galaxy
+```  
+  
+If your Ansible version >= 2.10.4, then  
+```yaml
+- hosts: localhost
+  vars:
+    galaxy_server_dir: /srv/galaxy
+    ansible_user: 'anyname'
+  connection: local
```
hexylena, line 332: We generally avoid encoding those in the playbook, they're part of the connection variables and should be written in the hosts file.

Areso: @hexylena 
I do understand you.
However:
1) connection: local was there before. I didn't change nor add it.
2) without the second parameter it fails and there is no mention that it should be set. So, we could drop all the cause, but somehow point it out, so a person will save some time and won't come to the chat for help with the already known issue.
It should be maximum user-friendly for newcomers, shouldn't it?

Areso: Will open a new PR, according to the notes.

hexylena:
There is no need to open a new PR, you can just update the existing one 
in the future please.

PR #122, opened 2021-01-19
Readme modify

Areso:
1) allows to avoid an error of implicit pointing to the galaxy
2) allows to avoid error, while downloading newer version from GitHub
@hexylena - made changes, please check it, is it OK now?

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ Install Galaxy on your local system with all the default options:
      - galaxyproject.galaxy
 ```  
   
-If your Ansible version >= 2.10.4, then you must indicate explicitly ansible_user in your hosts file, otherwise you will gen an error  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -320,9 +320,11 @@ Install Galaxy on your local system with all the default options:
...
   connection: local
   roles:
-     - galaxy
-```
-
+     - galaxyproject.galaxy
+```  
+  
+If your Ansible version >= 2.10.4, then you must indicate explicitly ansible_user in your hosts file, otherwise you will gen an error  
+
```
hexylena, line 327:
```suggestion
If your Ansible version >= 2.10.4, then when you run `ansible-playbook playbook.yml` you should supply an extra argument `-u $USER`, otherwise you will get an error.
  
```

hexylena:
This looks good, thanks for making those changes!

hexylena:
Thanks for documenting this here! I forget this document exists since we mostly refer people to the GTN materials.

PR #123, opened 2021-01-21
Supply data_dir instead of individual cache dirs

hexylena:
Per discussion in https://github.com/galaxyproject/ansible-galaxy/commit/a0e933dd1f6422dfca66b5680fd91a717ee52836, this could be a better option for us where we're less likely to miss something.

That said, we lose the nice "cache" distinction. @natefoo what do you think. This is obviously without any migration path, and I can guess how you feel about that, likewise I don't want the mutable dir to suddenly double in size due to 2x copies of cache data.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -179,18 +179,7 @@ galaxy_app_config_default:
   tool_sheds_config_file: "{{ galaxy_server_dir }}/config/tool_sheds_conf.xml.sample"
 
   # Various cache and temporary directories
-  template_cache_path: "{{ galaxy_cache_dir }}/template_cache"
-  whoosh_index_dir: "{{ galaxy_cache_dir }}/whoosh_cache"
-  object_store_cache_path: "{{ galaxy_cache_dir }}/object_store_cache"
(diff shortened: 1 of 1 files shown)


mvdbeek:
Setting `data_dir` could also be in addition to the specific path, it's not necessarily either/or

hexylena:
True! but we'll still maybe miss new cache-type directories. 

I was vaguely thinking maybe we need a `cache_dir` in galaxy, like `data_dir`, by default could be below data_dir, but would be for anything that could be freely deleted?

mvdbeek:
Sounds reasonable to me as a feature request.

hexylena:
I'll spec it out / file it. Cheers!

natefoo:
+1000, this was exactly the point of `data_dir`.

@mvdbeek:
> Setting `data_dir` could also be in addition to the specific path, it's not necessarily either/or

This change doesn't preclude people from setting the options directly in `galaxy_config.galaxy`, it just causes them not to be set by the role if unset by the user.

PR #126, opened 2021-03-02
Only create paths if they exist

rhpvorderman:
Some directories do not exist on the custom layout. I.e. have a `null` variable. (The local tool path I believe). And these directories can not be created. Using a `when: item | default(False)` will prevent these `null` directories from being included and created (which causes a crash).

### DIFF
--- a/tasks/paths.yml
+++ b/tasks/paths.yml
@@ -50,6 +50,7 @@
         group: "{{ __galaxy_privsep_user_group }}"
         mode: "{{ __galaxy_dir_perms }}"
       with_items: "{{ galaxy_privsep_dirs }}"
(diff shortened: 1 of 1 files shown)


rhpvorderman:
This line to be specific. https://github.com/galaxyproject/ansible-galaxy/blob/586b66562ae8e1d1b0ad43359c43f5a28eeec3fc/vars/layout-custom.yml#L12

If `galaxy_manage_paths: true` the task creating the paths will crash.
This patch fixes that by omitting `null` paths.

hexylena (approved):
looks fine to me.

rhpvorderman:
Thanks @hexylena!

PR #127, opened 2021-03-23
Allow the use of miniconda python instead of system python.

### LINKED ISSUE #110: Install miniconda and make virtualenv from conda
This would be pretty good to have for e.g. CentOS 7, where Python 3 may not be installed (or if it is, it may not be installed on the cluster). Having an option to use conda for Galaxy's venv/python would fix this.

rhpvorderman:
Simple changes to allow the user to use a python from a miniconda installer instead of system python. This is useful on systems where python is old.

Fixes #110 

I did also investigate options for reusing the existing miniconda installation as the galaxy dependency resolver, but that turned out to cause some major headaches when using privilege separation. Also, ideally, the environments are installed separate from the conda_root_prefix (not in root_prefix/envs  but in the mutable data directory under dependencies/_conda/envs) but I could not yet make this happen.
It is a lot of hassle and `conda_auto_init` works just fine on most systems.


hexylena:
I'm not sure if you've discussed with @natefoo at all, but we have https://github.com/galaxyproject/ansible-miniconda/, would it be maybe useful to use that rather than a custom conda.yml here? We could call that role twice, once for galaxy conda, once for job dependency conda maybe? And then just tell people to set `galaxy_virtualenv_python = [galaxy conda]`

rhpvorderman:
Whoops I did not know that existed. Sorry. I might have simply copied the tasks from there as there are obviously no copyright issues. 

It would be very useful to reuse roles in the way you mention.  It can theoretically be accomplished with include_role or import_role. But there is no way to signify to ansible that these roles need to be installed. There are role_dependencies, but these are automatically run before the role that depends on them. Given that this is an optional dependency it complicates things even further. The machinery will get quite complicated if you somehow manage to solve this.
This PR installs conda with 26 lines of yml (+3 for the include). It is quite straightforward. I think a  more modular solution based on ansible miniconda will result in increased complexity and not necessarily enhanced functionality. In the end it will replace 3 normal tasks with one include/import_role task and a lot of dependency headaches.

hexylena:
@natefoo wrote that role with the explicit intention of it being used for this purpose, so he may have some plans there :)

I think we have a couple options here though:

1. With the [galaxy ansible-nginx role](https://github.com/galaxyproject/ansible-nginx/search?q=nginx_ssl_role) we're already depending on an outside role in one place, I think we could do something similar here? There we let users specify a 'role url' that's used, and they're expected to ensure it's present themselves.
2. We could try something like this https://stackoverflow.com/questions/57581363/ansible-role-dependencies-install-but-dont-run-yet-how, between the post + the comments it seems we could do a `when: False` in the role dependencies section, and then maybe use an `include_role` stanza?

rhpvorderman:
Hmm it looks like it is possible in ansible 2.10: https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#using-meta-requirements-yml

So the roles will be installed, yet not executed. That would be a great solution.

rhpvorderman:
Allright. I see that this role is ansible 2.9 dependent. Are we willing to upgrade to enable this feature? It would be pretty neat!

I can make this happen. I want to make these changes to create a container that is slightly heavier than galaxy-k8s which can be used in docker-compose setups and galaxy test scripts. So I have some time I can allocate to this.

hexylena:
Fine from my side to upgrade, that sounds good!

tir. 23. mar. 2021, 13:50 skrev Ruben Vorderman ***@***.***>:

rhpvorderman:
I did some further investigations. Requirements work only for `ansible-galaxy install` not for `ansible-playbook` so users will have to install the dependencies first if they checkout ansible-galaxy from git. So it still is a lot of hassle to include roles modularly into other roles.
I am in favour of keeping the three-task yaml. All the extra functionality of the miniconda role is not necessary in this role.

hexylena:
> checkout ansible-galaxy from git

we don't support that, the only supported workflow is downloading it with ansible-galaxy. Does that simplify things?

rhpvorderman:
Sure. If we are okay with going from zero to one dependency that is. That is quite a big move for any project.

natefoo:
Thanks for the submission @rhpvorderman!

This role is already doing a ton and I worry about continuing to expand the scope. I don't think the role needs to do everything - for example, PostgreSQL is an optional (but not really optional) dependency for a functional Galaxy server and that is already split out into two separate roles. And that is for a dependency that essentially everyone will use with Galaxy - using Conda for the Galaxy Python will be less common.

I actually am not a huge fan of the way the nginx/certbot role dependency works but we have no choice in that case since you have to have a functioning non-SSL web server set up before you can request a cert with an HTTP-01 challenge, but the server will fail to start if you configure the SSL vhosts before the certs are in place - so the certbot role *must* be run in the middle of setting up nginx. But the certbot role is still optional and the nginx role instructions explain how to install the certbot role and configure the nginx role to use it if that is what you need.

In general all of the galaxyproject roles are designed to be as modularized and self-contained as possible to make them as flexible as people's various Galaxy deployments need to be. A playbook to install a full production Galaxy server is going to individually include at least 5 roles, probably more, so unless I misunderstand something, I don't think it should be a problem for people who want to use Conda as their Galaxy Python to run `galaxyproject.miniconda` before `galaxyproject.galaxy`?

rhpvorderman:
Ah okay. That makes sense. But in that case users can specifiy galaxy_venv_python themselves right?
So there is not really a point in adding this functionality or depending on ansible-miniconda then, am I right?

natefoo:
> So there is not really a point in adding this functionality or depending on ansible-miniconda then, am I right?

Right, you can just run the `galaxyproject.miniconda` role directly in your playbook, it does not need to be run from within this role.

> But in that case users can specifiy galaxy_venv_python themselves right?

I think you'd probably want to set `pip_virtualenv_command` to `{{ miniconda_prefix }}/envs/{{ galaxy_conda_env }}/bin/python3 -m venv` (or possibly `.../python3 -m venv` or just `.../virtualenv`).

rhpvorderman:
Thanks @natefoo . I gues this PR is unnecessary. What needs to be done with #110 then?

natefoo:
Probably just to update the GTN exercise to use `galaxyproject.miniconda`.

PR #129, opened 2021-04-12
Update setuptools as well as pip when creating a virtualenv.

rhpvorderman:
When a python from conda is used, this step prevents errors during the dependency installation phase.

This addition is necessary to work with conda python.


rhpvorderman:
Can this tiny change be merged in order to make everything work with conda python? @natefoo ?

natefoo:
Thanks @rhpvorderman, I tagged 0.9.13 and imported to Ansible Galaxy.

rhpvorderman:
Thanks!

PR #130, opened 2021-04-13
Make sure tool data is in the mutable data dir

rhpvorderman:
Where it should go according to the different layouts is debatable, but I opted to put it always in the mutable data dir except for the legacy layout where it is in galaxy_server_dir.


natefoo:
This overrides the default path in `<galaxy_root>/tool-data`, which could cause some warnings, but I don't think it should cause any significant problems. Thanks @rhpvorderman!

rhpvorderman:
Thanks @natefoo for merging! :smile:

PR #132, opened 2021-06-02
Add systemd, github actions support

hexylena:
I've commented out the support for zerglings due to lack of bandwidth to test (there are so many variables I'm probably missing something.)

We can always re-enable it later.


hexylena:
Well hot damn, after some misery it works.

tasks/systemd-zerg.yml
```diff
@@ -0,0 +1,78 @@
+---
```
natefoo, line 1: Maybe drop this and the zergling/zergpool unit files entirely? As we'll be moving away from uWSGI anyway, so are unlikely to ever put in the effort to make this work.

templates/systemd/galaxy-handler@.service.j2
```diff
@@ -0,0 +1,23 @@
+[Unit]
```
natefoo, line 1: Also drop since it's not handled in `galaxy_systemd`?

templates/systemd/resource_controls.conf.j2
```diff
@@ -0,0 +1,31 @@
+;;
```
natefoo, line 1: Not used anywhere?

natefoo (approved):
🎉

defaults/main.yml
```diff
@@ -274,3 +277,17 @@ galaxy_client_build_steps:
     - stageLibs
     - plugins
 
+
+# systemd options
+galaxy_restart_handler_name: "{{ 'galaxy mule restart' if (galaxy_systemd.mode == 'mule' and galaxy_manage_systemd) else 'default restart galaxy handler' }}"
+galaxy_systemd:
```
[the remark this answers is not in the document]
hexylena: bah humbug. ok

tasks/systemd-zerg.yml
hexylena: yeah, that's a very fair point.

templates/systemd/resource_controls.conf.j2
hexylena: we could, but, yeah, guess we won't now

hexylena:
Removed all the unused nonsense :) fingers crossed it passes

hexylena:
I'll click myself, so I can test the training. Feel free to revert if it was bad :|

PR #133, opened 2021-06-02
Support specification of ancillary configs in YAML

natefoo:
This is done in a way that should be fairly unobtrusive to people already specifying these configs via `galaxy_config_files`, `galaxy_config_templates`, and `galaxy_config.galaxy.*`.

For example, simply setting the new `galaxy_job_metrics_plugins` var in your vars will create `{{ galaxy_config_dir }}/job_metrics_conf.yml` with the var contents, and set the corresponding path option in `galaxy.yml` (unless overridden by explicitly setting `galaxy_config.galaxy.job_metrics_config_file`). The var syntax matches the Galaxy YAML plugin syntax (which we need to document and provide examples for by converting `job_metrics_conf.xml.sample`).

I believe all the following should be supportable via this mechanism since they all use the plugin config lib:

- [x] job metrics
- [x] dependency resolvers
- [x] container resolvers

The Pulsar role has a syntax that is converted to XML that should probably be switched to this syntax instead. See galaxyproject/ansible-pulsar#19 for discussion and [here for an example](https://github.com/galaxyproject/usegalaxy-playbook/blob/4c0479deffad7e94bd8a30eccf897e0a093128a8/env/common/host_vars/vm030.bridges2.psc.edu.yml#L70) of the existing Pulsar role syntax in action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,40 @@ Options for configuring Galaxy and controlling which version is installed.
 - `galaxy_clone_depth` (default: unset): Depth to use when performing git clone. Leave unspecified to clone entire
    history.
 
+**Additional config files**
+
+Some optional configuration files commonly used in production Galaxy servers can be configured from variables:
+
+- `galaxy_dependency_resolvers`: Populate the `dependency_resolvers_conf.yml` file. See the [sample XML
+  configuration][dependency_resolvers_conf_sample] for options.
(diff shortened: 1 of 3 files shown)


hexylena:
Love this! let's go with it

hexylena (approved):

PR #134, opened 2021-06-04
Move systemd vars to individual vars to make them easier to override

natefoo:
I also dropped the uWSGI stats port stuff - you can just set `stats: :port` in the `uwsgi` section of `galaxy.yml` to enable it if you want, and I didn't want to add more uWSGI-specific stuff to the role when we're deprecating uWSGI anyway.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -277,17 +277,27 @@ galaxy_client_build_steps:
     - stageLibs
     - plugins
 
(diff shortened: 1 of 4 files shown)


hexylena (approved):
this seems like a yuuuge improvement over my rubbish, cheers.

til you can do envs on multiple lines, that's way clearer.

natefoo:
Will be released as 0.9.15 shortly.

PR #135, opened 2021-06-07
Enable optional dynamic job rules as templates

pcm32:
This PR enables the use of dynamic job rules as ansible templates, distinguishing from pure python setups.


natefoo:
Thanks @pcm32, looks good! I updated the task names to be the same as what we use earlier just to be consistent.

abretaud:
Hi!
Looks like this PR broke our playbook on usegalaxy.fr: we have some yml files in `galaxy_dynamic_job_rules_src_dir` too, and now they are just ignored as they don't end with .py or .py.j2
I guess .eu is doing the same with putting yml files in this dir
Would you have time to fix it?

natefoo:
@abretaud sorry about that - how does the fix in #141 look?

PR #141, opened 2021-11-18
Fix sending dynamic rules files that do not end in .py

natefoo:
Reported by @abretaud in the comments of #135.


natefoo:
Yeah very strange, didn't encounter this locally. Python or Ansible version maybe?

abretaud:
So here's a fix for the strange syntax problem: https://github.com/galaxyproject/ansible-galaxy/pull/142
I also added a notify to restart galaxy when these dynamic rules are changed

abretaud:
Hum, not sure if adding a notify is needed: if `watch_job_rules` is enabled, is it supposed to reload the destinations everytime I modify a sorting hat yml file?

PR #142, opened 2021-12-10
Try to fix "found unknown escape character" in #141

abretaud:
following the advice in https://stackoverflow.com/questions/68637451/ansible-regex-search-unknown-escape-character


hexylena:
fascinating :upside_down_face:

Issue #140, opened 2021-10-25
Option to install additional python packages into virtual environment


cat-bro:
There could be a list variable `galaxy_additional_venv_packages` to specify additional python packages to install into the galaxy virtual environment.  My use case is that Galaxy Australia wants to install a package 'total-perspective-vortex' into galaxy's virtual environment.  The most logical time to install this than after the venv is created and before galaxy is restarted.

natefoo:
Fixed in #150, thanks @cat-bro!

PR #150, opened 2022-04-06
variable `galaxy_additional_venv_packages` to install extra python packages in galaxy venv

cat-bro:
For issue https://github.com/galaxyproject/ansible-galaxy/issues/140

The use case for Galaxy Australia is needing to install the `total-perspective-vortex` package into the venv after it has been created but before the restart handler runs.

### DIFF
--- a/tasks/dependencies.yml
+++ b/tasks/dependencies.yml
@@ -44,7 +44,7 @@
       environment:
         PYTHONPATH: null
         VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
-      when: galaxy_additional_venv_packages is not []
+      when: galaxy_additional_venv_packages
 
   remote_user: "{{ galaxy_remote_users.privsep | default(__galaxy_remote_user) }}"
   become: "{{ true if galaxy_become_users.privsep is defined else __galaxy_become }}"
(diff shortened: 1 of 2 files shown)


tasks/dependencies.yml
```diff
@@ -36,6 +36,16 @@
...
 
+    - name: Install additional packages into galaxy's virtual environment
+      pip:
+        name: "{{ galaxy_additional_venv_packages }}"
+        virtualenv: "{{ galaxy_venv_dir }}"
+        virtualenv_command: "{{ galaxy_virtualenv_command | default(pip_virtualenv_command | default(omit)) }}"
+      environment:
+        PYTHONPATH: null
+        VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+      when: galaxy_additional_venv_packages is not []
```
hexylena, line 47: maybe this check could be `when: galaxy_additional_venv_packages`, I think that should be falsey?

hexylena (commented):
Thanks for adding this, I've written something similar, but never contributed it back, awesome :)

tasks/dependencies.yml
cat-bro: Thanks!  I'm not sure why I thought that `when` would work.

Good CI btw!

hexylena:
fab!

PR #152, opened 2022-04-07
Check for both alembic and SQLAlchemy Migrate errors

### LINKED ISSUE #151: Database task breaks on dev
The database task breaks on dev (or anything past 22.01). The cause is that the task expects output generated by the SQLAlchemy Migrate tool, which is no longer present in dev, or 22.05.

I'll fix this.

ic4f:
Fixes #151 

This will support both pre- and post-alembic versions. No change to galaxy core is required.

### DIFF
--- a/tasks/database.yml
+++ b/tasks/database.yml
@@ -13,7 +13,7 @@
       failed_when:
         - current_db_version.rc != 0
         - "'migrate.exceptions.DatabaseNotControlledError' not in current_db_version.stderr"
-        - "'galaxy.model.migrations.IncorrectVersionError' not in current_db_version.stderr"
+        - "'galaxy.model.migrations.NoVersionTableError' not in current_db_version.stderr"
       when: not ansible_check_mode
 
     - name: Get maximum Galaxy DB version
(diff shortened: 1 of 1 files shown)


natefoo:
Will `galaxy.model.migrations.IncorrectVersionError` not be thrown if the DB is initialized but at the wrong version?

ic4f:
> Will `galaxy.model.migrations.IncorrectVersionError` not be thrown if the DB is initialized but at the wrong version?

That would be `OutdatedDatabaseError`. The `IncorrectVersionError` is only thrown [here](https://github.com/galaxyproject/galaxy/blob/6f3bdd39d90a7d2d729aa859f7dd67f3ba3a4212/lib/galaxy/model/migrations/__init__.py#L473) 
Is my logic faulty?

ic4f:
Don't merge just yet - checking whether a different error may be more appropriate here (on the galaxy side).

ic4f:
@natefoo this should be better. Will work correctly after https://github.com/galaxyproject/galaxy/pull/13695 is merged. 
Essentially, the previous version was correct: that *was* the error that was raised when the version table was not found (as is the case with a new db). However, it was the same error as the one raised in a different context - when the db is non-empty, is under SQLAlchemy Migrate control AND is outdated (pre-180). Now it raises a more appropriate error: NoVersionTable. This is handled by code that is triggered only when `scripts/manage_db.py` is called directly, bypassing the shell script `manage_db.sh` and passing SQLAlchemy Migrate-style command arguments (version/db_version) - which is why I didn't catch this use case earlier.

ic4f:
@natefoo does this look OK to you?

natefoo:
LGTM, thanks!

PR #154, opened 2022-05-01
fix a typo in the name of the option

gmauro:
the name of this option is without the s in the word "container"
https://github.com/galaxyproject/galaxy/blob/5846886edb176d072a5043c6abf1777d756fc5c0/lib/galaxy/config/sample/galaxy.yml.sample#L500


hexylena (approved):

natefoo:
Well it used to be containers_resolvers..., and I vaguely recalled we discussed changing it. [I guess someone did that](https://github.com/galaxyproject/galaxy/pull/12344) and I forgot about it. ;D

The new name works as of 21.09, this is probably ok. We could be fancy and use the version checks but that seems like overkill.

PR #164, opened 2022-07-19
[WIP] Temporary systemd units

hexylena:
To be replaced with gravity writing out systemd, asap. https://github.com/galaxyproject/gravity/issues/63


hexylena:
No longer needed, EU has their own fork where they're continuing to use pre-gravity style deployment.

PR #166, opened 2022-07-26
[WIP] Add a task to specify the brand color?

guerler:
This PR attempts to add an option to specify the brand color in the playbook vars. Suggestions on improving this are very welcome, particularly the fixed path to the css theme file seems to be questionable. Also wondering if and how tests could be added. It might make sense to integrate this as Galaxy config option instead. Thanks in advance.


guerler:
Closing this, it probably makes more sense to handle this within Galaxy, maybe with an additional configuration file.

PR #169, opened 2022-09-14
Add options for cleaning up jobs/tmp dir

natefoo:
We could add `pgcleanup.py` here as well.


hexylena:
Ahhh!!!! 😍😍😍😍 Beautiful

I **just** installed it on one of my servers yesterday due to 300GB worth of failed uploads being leftover (non tus server)

This is fantastic

PR #178, opened 2022-12-19
Adjust for recent gravity changes

guerler:
xref: #80. Thank you @natefoo for the suggested fixes. They did resolved the issue I ran into.

### DIFF
--- a/handlers/gravity_23.0.yml
+++ b/handlers/gravity_23.0.yml
@@ -6,9 +6,13 @@
   environment:
     GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
   listen: "restart galaxy"
+  become: yes
+  become_user: "{{ __galaxy_user_name }}"
 
 - name: galaxyctl update
   command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} update"
   environment:
     GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
   listen: "galaxyctl update"
+  become: yes
+  become_user: "{{ __galaxy_user_name }}"
--- a/handlers/gravity_23.0.yml
+++ b/handlers/gravity_23.0.yml
@@ -1,0 +1,14 @@
+---
+# handlers for gravity as used in galaxy version 23.0
+
+- name: galaxy gravity restart
+  command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} graceful"
+  environment:
+    GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
+  listen: "restart galaxy"
+
+- name: galaxyctl update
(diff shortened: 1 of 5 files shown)


natefoo:
Thanks! Unfortunately we'll need to make this conditional for the Galaxy (or Gravity) version since it's not backwards compatible with Gravity 0.x.

handlers/gravity_23.0.yml
```diff
@@ -0,0 +1,18 @@
+---
+# handlers for gravity as used in galaxy version 23.0
+
+- name: galaxy gravity restart
+  command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} graceful"
+  environment:
+    GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
+  listen: "restart galaxy"
+  become: yes
+  become_user: "{{ __galaxy_user_name }}"
```
natefoo, line 10: This would fail on Main since the playbook connects as a different unprivileged user that can't sudo to other users. That said, in my case I have disabled the role handlers by setting `galaxy_systemd_mode` to `null` and defined my own, so maybe that is sufficient.

mvdbeek: There are some different ways in this role to do that, is there one that would work ?

tasks/gravity.yml
```diff
@@ -8,11 +8,19 @@
...
     - name: Update Gravity process management files
       command: "{{ galaxy_venv_dir }}/bin/galaxyctl update"
       args:
         creates: "{{ galaxy_gravity_state_dir }}/supervisor/supervisord.conf.d"
+      when: __galaxy_major_version is version('23.0', '<')
+
+    - name: Update Gravity process management files
+      command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} update"
+      args:
+        creates: "{{ galaxy_gravity_state_dir }}/supervisor/supervisord.conf.d"
```
hexylena, line 22: this still looks supervisord specific, not sure if we need to do something else here under systemd?

mvdbeek: Should all be stateless at this point, I think we can just delete this

mvdbeek: actually i think the `galaxy_systemd_mode == gravity` case is still the default using supervisor and all, so this is probably correct ?

hexylena: Do we need a separate `galaxy_systemd_mode == gravity-systemd` then or similar?

Judging by https://github.com/galaxyproject/ansible-galaxy/blob/main/defaults/main.yml#L383 there just isn't support for the systemd version of gravity in here yet and that might need more changes (unless i'm misreading)

natefoo: It just depends on your Gravity config, there shouldn't be anything pm-specific.

Yes it's stateless now but supervisor mode still has to create supervisor PMs and those go in the "state dir" (which is now inside Galaxy's `data_dir` unless you explicitly set the Gravity state dir).

PR #182, opened 2023-03-06
Drop modifying venv activate to set `$GRAVITY_STATE_DIR`

natefoo:
This was only ever done as a convenience and is no longer needed or desirable with Gravity 1.0 (keeping it generally won't harm anything either, although it might cause confusion down the line).


natefoo:
@jdavcs is this an expected change to `manage_db.py` in dev that we need to adjust the role for?

jdavcs (approved):

PR #183, opened 2023-03-09
Fix exception fully qualified name

jdavcs:
Exception classes were moved into their own module in [#15509 (galaxy)](https://github.com/galaxyproject/galaxy/pull/15509/commits/df0c6b629243d4b897611d096423cfcee37d91e2).


jdavcs:
No, that's not it.

natefoo:
The test failure is unrelated and caused by Node switching their binary build platform for Node v18 to EL8. This will be a problem going forward as we don't want to drop support for El7 yet. I spoke with @dannon who verified that the client still builds with v16, so I'll add a conditional in a separate PR that forces v16 on EL7 when Galaxy >= 23.0.

natefoo (approved):
Thanks @jdavcs!

PR #184, opened 2023-03-09
Support defining a maximum node version for a given platform

natefoo:
As discussed in #183.


jdavcs (approved):
lgtm

jdavcs:
@natefoo How did you reproduce it locally? (i.e., did you start galaxy? also, what's your database setup - I assume at least one db is sqlite?) From what I saw in the molecule output, I guessed it was a gravity restart one-of error - did I misread that log? Or was the migrations error visible only only locally?

PR #185, opened 2023-04-06
Updates for systemd mode Gravity

natefoo:


natefoo:
Test failures due to Mercurial version update in dev, retest after https://github.com/galaxyproject/wheelforge/pull/19

hexylena (approved):
Nice, like the removal of the special variable, and instead relying on galaxy.yml

PR #189, opened 2023-04-13
Fix directory creation

mira-miracoli:


hexylena (approved):
Thanks @mira-miracoli (and @mvdbeek)

PR #191, opened 2023-04-13
Add variables for creating extra dirs

natefoo:
Revert the changes in #188, #189, and #190 since this resulted in incorrect permissions. At some future point maybe we can get these right in an automated way, but for now, just add vars for adding to the dirs that already get precreated by the paths tasks.


hexylena (approved):
Classic, I needed to be humbled

natefoo:
Wasn't you, I started us down the wrong path

PR #192, opened 2023-04-14
Allow changing mode of static files

hexylena:
for e.g. welcome.html to be readable by nginx

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -246,6 +246,7 @@ galaxy_app_config_section: "{{ 'galaxy' if galaxy_config_style in ('yaml', 'yml'
 
 # Galaxy configuration files will be written with these permissions (mode argument to Ansible copy/template module)
 galaxy_config_perms: 0640
+galaxy_config_perms_public: 0644
 
 # The default Galaxy configuration, ensures that Galaxy can find all of the configs if galaxy_config_dir !=
 # galaxy_server_dir
@@ -313,6 +314,8 @@ galaxy_errordocs_prefix: /error
 # templates to be installed on the managed host.
 galaxy_config_files: []
 galaxy_config_templates: []
+# Like above, except with world readable permissions
+galaxy_config_files_public: []
 
 # Default Gravity configuration
 galaxy_gravity_state_dir: "{{ (galaxy_mutable_data_dir, 'gravity') | path_join }}"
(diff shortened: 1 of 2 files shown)


hexylena:
```
148 ubuntu@gat-8:~/git-gat$ ls -al files/galaxy/themes.yml 
-rw-rw-r-- 1 ubuntu ubuntu 1654 Apr 14 11:55 files/galaxy/themes.yml
ubuntu@gat-8:~/git-gat$ ls -al /srv/galaxy/config/themes.yml
-r----xrw- 1 root galaxy 1654 Apr 14 11:56 /srv/galaxy/config/themes.yml
```

yeah that's relevant. And wow what bizarre permissions. 

I'm guessing because internally we have an octal number (which yaml parses as a number) for `galaxy_config_perms`, 

```
>>> yaml.load("test: 0640")
{'test': 416}
```

yeah that's that nonsense for sure.

hexylena:
This is...probably better.

kysrpex:
Thanks for fixing this, I was using a post task up to now.

hexylena:
@kysrpex which files were you using it for? (i.e. is there something in the training that should have more readable permissions?)

natefoo:
How is it not parsing the defaults as octal? I'm never clear on when it does and doesn't. But seems like we should just quote both of them to be explicit that they are not octal?

natefoo (approved):

hexylena:
> How is it not parsing the defaults as octal?

It does, directly into an int (416). So if that gets turned into a string (as happens via `"{{ ... }}"`) then it gets mis-interpreted. As far as I can tell there's no way to tell ansible "hey this is actually an int even though you wanted it quoted because it's a templated var".

Which means it's probably... impossible?

> I'm never clear on when it does and doesn't. But seems like we should just quote both of them to be explicit that they are not octal?

`0NNN` (parse by yaml), `"0NNN"` (parsed by ansible) both get turned into octals, `"NNN"` also for some reason gets interpreted as octal, rather than anything else. from the docs:

> For those used to /usr/bin/chmod remember that modes are actually octal numbers. You must either add a leading zero so that Ansible's YAML parser knows it is an octal number (like 0644 or 01777)or quote it (like '644' or '1777') so Ansible receives a string and can do its own conversion from string into number. Giving Ansible a number without following one of these rules will end up with a decimal number which will have unexpected results.

PR #193, opened 2023-04-18
Fix issue with jinja2 < 2.11.0

nsoranzo:
Ubuntu 20.04 LTS has jinja2 2.10.1, which doesn't support `is true` and various other `is` tests added in https://github.com/pallets/jinja/pull/824

Fix:
```
TASK [galaxyproject.galaxy : Schedule tmpclean cron job (root)] *******************************************************************************************************************************************
fatal: [gat-18.eu.galaxy.training]: FAILED! => 
  msg: |-
    An unhandled exception occurred while templating '{{ (galaxy_tmpclean_log is not true) | ternary(
        ((galaxy_tmpclean_log is none) | ternary(
            '>/dev/null',
            '>>' ~ galaxy_tmpclean_log
        )),
        ''
    ) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: template error while templating string: no test named 'true'. String: {{ (galaxy_tmpclean_log is not true) | ternary(
        ((galaxy_tmpclean_log is none) | ternary(
            '>/dev/null',
            '>>' ~ galaxy_tmpclean_log
        )),
        ''
    ) }}
```

when using ansible 2.12.10 .

Alternatively we could make clear that jinja >= 2.11.0 is required.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -188,7 +188,7 @@ _galaxy_tmpclean_command:
 
 galaxy_tmpclean_verbose_statement: "{{ galaxy_tmpclean_verbose | ternary(' -v', '') }}"
 galaxy_tmpclean_log_statement: >-
-  {{ (galaxy_tmpclean_log is not true) | ternary(
+  {{ (galaxy_tmpclean_log != true) | ternary(
       ((galaxy_tmpclean_log is none) | ternary(
           '>/dev/null',
           '>>' ~ galaxy_tmpclean_log


hexylena:
20.04 is quite old, hmm.

natefoo:
To be clear, this is with Ansible installed from `apt` on 20.04? Ansible itself appears to require unpinned jinja2, so if you `pip install 'ansible<2.13'` I think you'd get a version of jinja2 that supported this.

I am not sure when `foo == true` syntax will be dropped but IIRC it warns already.

nsoranzo:
No, it's a deb package from https://launchpad.net/~ansible/+archive/ubuntu/ansible .

natefoo (approved):
I seem to have been wrong about it warning... and the `!=` syntax for booleans is not deprecated?

PR #194, opened 2023-08-03
Themes

mira-miracoli:
I included the ansible-galaxy-subdomains role in this one and updated it to the new galaxy-integrated *per_host* functionality. See
- https://github.com/usegalaxy-eu/ansible-galaxy-subdomains/pull/11/
For details


mira-miracoli:
I think the failing tests might be caused by broken dependencies in dev

defaults/main.yml
```diff
@@ -23,6 +23,9 @@ galaxy_manage_gravity: "{{ false if __galaxy_major_version is version('22.05', '
 galaxy_manage_systemd: no # For Galaxy
 galaxy_manage_systemd_reports: no # For Reports
 galaxy_manage_cleanup: no
+galaxy_manage_themes: no
+galaxy_manage_static: no
```
hexylena, line 27: just catching up, this is semantically different than `galaxy_manage_static_setup`?  They sound so similar, I'm wondering if they shouldn't be the same variable?

tasks/setup_static_dirs.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: Create static dirs
+  ansible.builtin.file:
+    state: directory
+    mode: '0755'
+    owner: "{{ __galaxy_privsep_user_name }}"
+    group: "{{ __galaxy_privsep_user_group }}"
+    path: "{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}"
+
+- name: Move all contents from static to parent directory
+  ansible.builtin.command: cp -r {{ galaxy_themes_static_dir }}/. {{ galaxy_themes_static_path }}/static-{{ \
```
hexylena, line 11: what's the reason we don't use a copy task here?

hexylena (commented):
I am very excited for this!

hexylena:
@natefoo can you give this a look?

tasks/setup_static_dirs.yml
mira-miracoli: I'm not sure now. I think I wanted to make sure it works recursively, but that also work with the `directory_mode` in the copy module. I can test it

defaults/main.yml
mira-miracoli: Maybe the naming is a bit bad here. The variable enables the creation of static subdomain directories

mira-miracoli: I changed it, maybe this is better @hexylena ?

tasks/setup_static_dirs.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- name: Move all contents from static to parent directory
+  ansible.builtin.copy: 
+    src: "{{ galaxy_themes_static_dir }}"
+    dest: "{{ galaxy_themes_static_path }}/static-{{ \
+      subdomain.name }}"
+    owner: "{{ __galaxy_privsep_user_name }}"
+    group: "{{ __galaxy_privsep_user_group }}"
+    directory_mode: true
+  changed_when: false
```
kysrpex, line 18:
```suggestion
```

Do not forget to remove this, now `ansible.builtin.copy` determines whether something changed or not.

README.md
```diff
@@ -105,6 +105,40 @@ The handler should "listen" to the topic `"restart galaxy"`.
...
+│   ├── static
+│   │   ├── dist (optional)
+│   │   │   ├── another-static-image.svg
+│   │   │   └── more-static-content-2.svg
+│   │   └── welcome.html
+│   └── themes
+│       └── <subdomain-name-1>.yml
+... (and many more subdomains)
+~~~
+Were the <subdomain-name-1> should exactly match your subdomain's name. The subdirectories `static` and `themes` are mandatory, as well as the correctly named theme file (if you enabled `galaxy_manage_themes`), while all subdirectories in `static` are optional.
```
natefoo, line 132:
```suggestion

Where the <subdomain-name-1> should exactly match your subdomain's name. The subdirectories `static` and `themes` are mandatory, as well as the correctly named theme file (if you enabled `galaxy_manage_themes`), while all subdirectories in `static` are optional.  
```

defaults/main.yml
```diff
@@ -425,3 +472,64 @@ galaxy_systemd_env: []
...
+#     - "annotation"
+#     - "genome_diversity"
+#     - "multiple_alignments"
+#   extra_tool_labels:
+#     - "proteomics"
+galaxy_themes_static_keys:
+  static_dir: ""
+  static_images_dir: "images/"
+  static_scripts_dir: "scripts/"
+  static_welcome_html: "welcome.html/"
```
natefoo, line 493:
```suggestion
  static_welcome_html: "welcome.html"
```

tasks/copy_static_files.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Copy static files
```
natefoo, line 2:
```suggestion
- name: Copy subdomain static files
```

Also can this be renamed to something like `tasks/copy_subdomain_static_files.yml`?

tasks/setup_static_dirs.yml
```diff
@@ -0,0 +1,54 @@
...
+- name: Move all contents from static to parent directory
+  ansible.builtin.copy: 
+    src: "{{ galaxy_themes_static_dir }}"
+    dest: "{{ galaxy_themes_static_path }}/static-{{ \
+      subdomain.name }}"
+    owner: "{{ __galaxy_privsep_user_name }}"
+    group: "{{ __galaxy_privsep_user_group }}"
+    directory_mode: true
+
+- name: Remove static dirs
```
natefoo, line 19:
```suggestion
- name: Remove subdomain static dirs
```

defaults/main.yml
hexylena: i guess it makes sense to have these as vars even if I don't love them being overridden

defaults/main.yml
```diff
@@ -295,6 +298,50 @@ galaxy_app_config_default:
...
   visualization_plugins_directory: "config/plugins/visualizations"
 
+  # Static and themes configuration, will only be added if galaxy_manage_themes
+  enable_static: "{{ galaxy_manage_subdomain_static }}"
+  static_dir_by_host: >
+    { {% if galaxy_manage_subdomain_static %}
+    {% for subdomain in galaxy_themes_subdomains %}
+    '{{ subdomain.name }}.{{ galaxy_themes_instance_domain}}': '{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}/'
+    {% endfor %}
+    {% endif %} }
```
hexylena, line 308: I'm looking at alternative implementations here for what we can do.

I was thinking perhaps a solution here is to separate these out and leverage ansible's resolution process for variables. Here we could write the following (we should probably do it anyway just for cleanliness but).

```suggestion
  static_dir_by_host: {{ __static_dir_by_host }}
```

but ansible's `set_fact` module considers variable names with a `.` as invalid, they're literally forbidden whcih is wild to me.

```
---
- hosts: all
  vars:
    galaxy_themes_static_keys:
      static_dir: ""
      static_images_dir: "images/"
      static_scripts_dir: "scripts/"
      static_welcome_html: "welcome.html"
      static_favicon_dir: "favicon.ico"
      static_robots_txt: "robots.txt"
  pre_tasks:
    - find:
        paths: files/galaxy/static/
        patterns: "{{ galaxy_themes_static_keys.static_welcome_html }}"
        recurse: yes
      register: python_files

    - debug:
        msg: "{{ item.path }}"
      with_items: "{{ python_files.files }}"

    - set_fact:
        "{{ item.path | regex_replace('files/galaxy/static/', '') | regex_replace('/static/.*', '')}}": "{{ item.path }}"

      with_items: "{{ python_files.files }}"
      # with_fileglob: 'files/galaxy/static/*/static/{{ galaxy_themes_static_keys.static_welcome_html }}'


    - debug:
        msg: "{{ __static_dir_by_host }}"
```

so that's out. I tried a couple different ways including `from_yaml` to accomplish this, but, none worked.

It would be a *lot* easier if we could pull this information from another file, rather than sticking it all in galaxy.yml. @natefoo is there any good way to accomplish that? the only other option is appending to the end of galaxy.yml manually which is ,,, no. I don't want that.

hexylena: OK it works with from_json, ALRIGHT.

hexylena: <details><summary>Ignore this version I was dumb</summary>

```
    - find:                                                               
        paths: files/galaxy/static/                                       
        patterns: "{{ galaxy_themes_static_keys.static_welcome_html }}"   
        recurse: yes                                                      
      register: python_files
     # This is alternatively just `galaxy_themes_subdomains`, but I did dynamic discovery for some dumb reas.

    - set_fact:                                                                                                         
        __static_dir_by_host: |                                                                                         
          {{ '{' }}                                                                                                     
          {% for item in python_files.files %}                                                                          
            "{{ item.path | regex_replace('files/galaxy/static/', '') | regex_replace('/static/.*', '')}}": "{{ item.path }}",
          {% endfor %}                                                                                                  
          {{ '}' }}
    - debug:                                                   
        msg: "{{ __static_dir_by_host }}"                      
                                                               
    - debug:                                                   
        msg: "{{ __static_dir_by_host['test.example.com'] }}"                                  
```
</details>


yeah this works:

```
    - set_fact:                                                                                                         
        __static_dir_by_host: |                                                                                         
          {{ '{' }}                                                                                                     
          {% for subdomain in galaxy_themes_subdomains %}                                                                          
            "{{ subdomain.name }}.{{ galaxy_themes_instance_domain}}": "{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}/images",
          {% endfor %}                                                                                                  
          {{ '}' }}
    - debug:                                                   
        msg: "{{ __static_dir_by_host }}"                      
                                                               
    - debug:                                                   
        msg: "{{ __static_dir_by_host['test.example.com'] }}"                                  
```

I'm not sure I understand why that works but it works pretty OK I think.

We should write a `tasks/subdomain_discovery.yml` that sets these `__` variables ahead of computation (we should set defaults of `__static_dir_by_host: {}`, and then override them.), a set of tasks only included conditionally on `galaxy_manage_subdomains`, and then whenever galaxy.yml gets around to being read, it'll read those variables correctly.

hexylena (commented):
Ok, I think i've found a workaround for setting the variables, this is doable!

defaults/main.yml
mira-miracoli: Thank you so much for finding a solution :heart: 
Even though I don't understand why this works, either.
So we have to do this for each static_x_by_host thing, right? (images, robots, etc.)

defaults/main.yml
mira-miracoli: `welcome.html` is actually a directory

tasks/setup_static_dirs.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: Create static dirs
```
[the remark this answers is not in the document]
mira-miracoli: yes sure, thats less confusing

defaults/main.yml
hexylena: D: what

defaults/main.yml
hexylena: yep, one for each! Then we should be .. ready to go? exciting!

mira-miracoli: I can add this today, but there is one more thing, unfortunately – the static_dir copying task is not idempotent unfortunately. It is done each time you run the role. Maybe we could solve this by using the rsync module?
Problem there is that it takes the same directory name, which is a problem if you want to "clone" the directory in the same spot

defaults/main.yml
natefoo: wat

natefoo:
The archive/unarchive is going to result in a lot of always-run tasks, I think? But I also recall you mentioned some reason that using the `synchronize` module won't work?

mira-miracoli:
Are you ok with the latest changes?

defaults/main.yml
```diff
@@ -425,3 +486,64 @@ galaxy_systemd_env: []
...
+#     - "annotation"
+#     - "genome_diversity"
+#     - "multiple_alignments"
+#   extra_tool_labels:
+#     - "proteomics"
+galaxy_themes_static_keys:
+  static_dir: ""
+  static_images_dir: "images/"
+  static_scripts_dir: "scripts/"
+  static_welcome_html: "welcome.html/"
```
hexylena, line 507:
```suggestion
  static_welcome_html: "welcome/"
```

maybe we can remove the extension to make it more clear it's a folder, I was reviewing this again and got confused again

tasks/setup_static_dirs.yml
```diff
@@ -0,0 +1,51 @@
...
+    mode: '0755'
+    owner: "{{ __galaxy_privsep_user_name }}"
+    group: "{{ __galaxy_privsep_user_group }}"
+    path: "{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}/welcome.html"
+
+- name: Check if iframe for subdomain exists
+  ansible.builtin.uri:
+    url: "{{ galaxy_themes_welcome_url_prefix }}{{ subdomain.name }}.html"
+    return_content: true
+  register: iframe
```
hexylena, line 34:
```suggestion
  register: galaxy_themes_use_iframe
```

templates/welcome.html.j2
```diff
@@ -0,0 +1,39 @@
...
+		}
+	</style>
+</head>
+
+<body style="margin:0px">
+	<iframe id='center-iframe' src="" height="100%"
+		onload="this.style.position='static'; this.style.visibility='visible';"></iframe>
+	<script type="text/javascript">
+		var pseudo_rand = parseInt(Date.now() / 1000 / 60 / 10);
+		{% if iframe is defined %}
```
hexylena, line 31:
```suggestion
		{% if galaxy_themes_use_iframe is defined %}
```

README.md
```diff
@@ -105,6 +105,44 @@ The handler should "listen" to the topic `"restart galaxy"`.
...
+│   │   │   └── more-content.jpg
+│   │   └── welcome.html
+│   └── themes 
+│       └── <subdomain-name-1>.yml           
+├── <subdomain-name-2>                            
+│   ├── static
+│   │   ├── dist (optional)
+│   │   │   ├── another-static-image.svg
+│   │   │   └── more-static-content-2.svg
+│   │   └── welcome.html
```
hexylena, line 130:
```suggestion
│   │   └── welcome.html (optional)
```

README.md
```diff
@@ -105,6 +105,44 @@ The handler should "listen" to the topic `"restart galaxy"`.
...
+
+~~~bash
+files/galaxy/static
+├──<subdomain-name-1>
+│   ├── static
+│   │   ├── dist (optional)
+│   │   │   └── some-image.png 
+│   │   ├── images (optional)
+│   │   │   └── more-content.jpg
+│   │   └── welcome.html
```
hexylena, line 122:
```suggestion
│   │   └── welcome.html (optional, galaxyproject.org will be displayed otherwise.)
```

defaults/main.yml
```diff
@@ -23,6 +23,10 @@ galaxy_manage_gravity: "{{ false if __galaxy_major_version is version('22.05', '
 galaxy_manage_systemd: no # For Galaxy
 galaxy_manage_systemd_reports: no # For Reports
 galaxy_manage_cleanup: no
+galaxy_manage_themes: no
+galaxy_manage_subdomain_static: no
+galaxy_manage_host_filters: no
+galaxy_auto_brand: no # automatically sets the subdomain name as brand
```
hexylena, line 29: oh dang that's useful

tasks/main.yml
```diff
@@ -160,3 +160,56 @@
...
+    state: directory
+    mode: '0755'
+    owner: "{{ __galaxy_privsep_user_name }}"
+    group: "{{ __galaxy_privsep_user_group }}"
+    path: "{{ galaxy_themes_static_path }}/static/welcome.html"
+  when: galaxy_manage_subdomain_static
+
+- name: Template welcome.html for basedomain
+  ansible.builtin.template:
+    src: welcome.html.j2
```
hexylena, line 175: I think all of the `when: galaxy_manage_subdomain_static` could be moved into a separate file to match the rest of `main.yml` that does conditional includes.

hexylena (approved):
This looks fantastic and much more idemopotent! A couple nitpicky comments but this feels basically ready to ship

defaults/main.yml
mira-miracoli: but the folder is named automatically by Galaxy, I think removing this will break it.
https://github.com/galaxyproject/galaxy/blob/41b302de53e2d256dedefbdefd65f3709b6d85ee/lib/galaxy/webapps/base/webapp.py#L1186
https://github.com/galaxyproject/galaxy/blob/41b302de53e2d256dedefbdefd65f3709b6d85ee/lib/galaxy/web/framework/middleware/static.py#L34

hexylena: Ohhh :skull:

Ok, guess we can't change that.

hexylena:
🎉

kysrpex:
> 🎉

This was a long one, congrats to you all :rocket:!

PR #198, opened 2023-12-01
Symlinks reduce disk usage

mira-miracoli:
This is a small improvement for the copy tasks in the static directories
As the role currently works, the whole static directory is `rsync`ed into every new subdomain directory, which leads to additional disk usage of ~ 154 MB for each subdomain in a basic (GTN) Galaxy setup.

With this improvement, several static subdirectories are symlinked and all files inside `dist` are symlinked, which reduces the disk usage per subdomain to ~ 2.3 MB.

However the disadvantage of this is:
- more extensive logs; each symlink needs to be checked and is logged by ansible, around 200 per subdomain, the `galaxy_themes_no_symlink_logs` just reduces the extent, but ansible still prints one line per file and subdomain
- slightly slower playbook runs; `synchronize` for the whole folder is faster than invoking an ansible operation for each file and we don't get rid of the `synchronize` fully, because we would need to create multiple subdirectories manually and iterate over them with symlink tasks

To solve this conflict, I added variables that let the user decide weather to use the symlink feature and reduce the logs or not

### DIFF
--- a/tasks/static_subdomain_dirs.yml
+++ b/tasks/static_subdomain_dirs.yml
@@ -5,7 +5,40 @@
     mode: '0755'
     owner: "{{ __galaxy_privsep_user_name }}"
     group: "{{ __galaxy_privsep_user_group }}"
-    path: "{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}"
+    path: "{{ galaxy_themes_static_path }}/{{ item }}"
+  with_items:
+    - "static-{{ subdomain.name }}"
+    - "static-{{ subdomain.name }}/dist"
+
+- name: Register files in static/dist
+  ansible.builtin.find:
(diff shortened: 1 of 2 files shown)


tasks/static_subdomain_dirs.yml
```diff
@@ -5,7 +5,40 @@
...
+    dest: "{{ galaxy_themes_static_path }}/static-{{ subdomain.name }}/{{ item }}"
+    state: link
+  when: galaxy_themes_symlinks
+  ignore_errors: true
+  with_items:
+    - "plugins"
+    - "patmat"
+    - "toolshed"
+    - "wymeditor"
+    - "style"
```
natefoo, line 41: This may end up being cumbersome to maintain, especially across changing Galaxy versions, but I reckon we can deal with that when it becomes a problem.

natefoo (approved):

PR #201, opened 2024-01-16
Fixes #200

### LINKED ISSUE #200: Automatic addition of `themes_config_file_by_host: {}` disables themes
After updating this role to version 0.10.19 our themes have disappeared, along with the option to select a theme in user preferences.

In our galaxy.yml file we have

```
    themes_config_file: /mnt/galaxy/config/themes_conf.yml
    themes_config_file_by_host: {}
```

We can get our themes back by commenting out `themes_config_file_by_host`.

mira-miracoli:
I tested this with my test instance, but please check if there could be scenarios where we don't want to remove configuration keys with length = 0

I nominate this for the ugliest jinja2 solution

### DIFF
--- a/templates/galaxy.yml.j2
+++ b/templates/galaxy.yml.j2
@@ -25,4 +25,7 @@
 {% for section, val in galaxy_config | dictsort if section not in ('uwsgi', 'gravity', galaxy_app_config_section) %}
 {{ {section: val} | to_nice_yaml(indent=4) }}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


hexylena (approved):
This makes sense to me, clean out empty mapping keys because they are for some horrid reason semantically meaningfully different from an unset one

mira-miracoli:
Just to be sure (even if I tested on my test instance) @cat-bro is this fix working for you?

cat-bro:
It is, thanks @mira-miracoli !

cat-bro (approved):

PR #206, opened 2024-02-22
Support installing local tools into sections

natefoo:
This alters the previous behavior slightly, tools that do not specify a section will be placed in a section named **Local Tools** whereas previously they would be outside of a section.

The old simple syntax is still supported:

```yaml
galaxy_local_tools:
  - foo.xml
  - bar.xml
```

Is the same as:

```yaml
galaxy_local_tools:
  - file: foo.xml
    section_name: Local Tools
  - file: bar.xml
    section_name: Local Tools
```


hexylena (approved):
"sure let's just add this tiny feature for admin training" "oh no everyone is using it"

thanks! this is a nice change, i need to start using this at work.

microfuge:
This is super convenient! Thanks!

Issue #202, opened 2024-01-18
Gravity 23.0 handlers skipped


mtangaro:
Hi,
I was running the galaxyproject.galaxy role. After few tries I still get the handlers for starting galaxy skipped:

```
…
RUNNING HANDLER [galaxyproject.galaxy : galaxy gravity restart] **************************************************************************************************************************************************************************************
skipping: [localhost]
RUNNING HANDLER [galaxyproject.galaxy : galaxyctl update] ********************************************************************************************************************************************************************************************
skipping: [localhost]
PLAY RECAP *******************************************************************************************************************************************************************************************************************************************
localhost                  : ok=42   changed=14   unreachable=0    failed=0    skipped=33   rescued=0    ignored=0   
```

Minimal step to reproduce it:

group_vars/galaxy.yml
```
# Galaxy
galaxy_create_user: true # False by default, as e.g. you might have a 'galaxy' user provided by LDAP or AD.
galaxy_separate_privileges: true # Best practices for security, configuration is owned by 'root' (or a different user) than the processes
galaxy_manage_paths: true # False by default as your administrator might e.g. have root_squash enabled on NFS. Here we can create the directories so it's fine.
galaxy_layout: root-dir
galaxy_root: /srv/galaxy
galaxy_user: {name: "{{ galaxy_user_name }}", shell: /bin/bash}
galaxy_commit_id: release_23.0
galaxy_force_checkout: true
miniconda_prefix: "{{ galaxy_tool_dependency_dir }}/_conda"
miniconda_version: 23.9
miniconda_channels: ['conda-forge', 'defaults']
```

Playbook:
```
---
- hosts: galaxy
  connection: local
  become: true
  become_user: root
  pre_tasks:
    - name: Install Dependencies
      package:
        name: ['acl', 'bzip2', 'git', 'make', 'python-psycopg2', 'python3-psycopg2', 'tar', 'pip',  'gcc', 'openssl-devel', 'bzip2-devel', 'libffi-devel', 'zlib-devel', 'xz-devel', 'sqlite-devel', 'wget']
  roles:
    - role: galaxyproject.galaxy
      galaxy_build_client: false
```
Role: galaxyproject.galaxy,0.10.14

The client build is skipped to save a little bit of time during my tests. Actually I run the full GTN galaxy installation with ansible and in the end I get the handlers skipped.

[...]

PR #211, opened 2024-04-15
Update best practice in README.md

dometto:
Hi all,

I tried following the best practices guide in the README in the playbook I'm working on and noticed I had to deviate from it at some points. Maybe I did something wrong, but wanted to open this PR in case the best practice indeed needs to be updated. I'm pretty sure about the following changes:

* update to the postgres_objects role name
* `supervisorctl` should not be used

I needed to add the `systemd` `post_tasks` to actually get Galaxy to start up the first time, but maybe there's a better way for this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,7 @@ Install Galaxy as per the current production server best practices:
     galaxy_force_checkout: true
     galaxy_create_user: yes
     galaxy_manage_paths: yes
+    galaxy_manage_systemd: yes
     galaxy_user: galaxy
     galaxy_privsep_user: gxpriv
     galaxy_group: galaxy
@@ -429,21 +430,6 @@ Install Galaxy as per the current production server best practices:
       become: yes
       become_user: postgres
     - role: galaxyproject.galaxy
-  handlers:
-    - name: Galaxy gravity restart
-      command: "/usr/local/bin/galaxyctl graceful"
-      listen: "restart galaxy"
-  post_tasks:
-
-    - name: Get Galaxy service Status
-      ansible.builtin.systemd:
-        name: "galaxy.target"
-      check_mode: true
-      register: galaxy_status
-
-    - name: Galaxy gravity start
-      command: "/usr/local/bin/galaxyctl start"
-      when: "galaxy_status.status.ActiveState == 'inactive'"
 ```
 
 License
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -425,16 +425,25 @@ Install Galaxy as per the current production server best practices:
...
       become_user: postgres
     - role: galaxyproject.galaxy
   handlers:
-    - name: Restart Galaxy
-      supervisorctl:
-        name: galaxy
-        state: restarted
-      listen: restart galaxy
+    - name: Galaxy gravity restart
+      command: "/usr/local/bin/galaxyctl graceful"
```
hexylena, line 434:
```suggestion
      command: "{{ galaxy_gravity_command }} graceful"
```

README.md
```diff
@@ -425,16 +425,25 @@ Install Galaxy as per the current production server best practices:
...
+  post_tasks:
+
+    - name: Get Galaxy service Status
+      ansible.builtin.systemd:
+        name: "galaxy.target"
+      check_mode: true
+      register: galaxy_status
+
+    - name: Galaxy gravity start
+      command: "/usr/local/bin/galaxyctl start"
```
hexylena, line 445:
```suggestion
      command: "{{ galaxy_gravity_command }} start"
```

hexylena:
> in case the best practice indeed needs to be updated.

thank you for opening this, yes, they do. We haven't been keeping the readme up to date, not many people are reaching this repository via the readme and trying to setup galaxy with just that information. Most folks [go through the course](https://gxy.io/GTN:P00000) where we document the 'best practices' more thoroughly.

README.md
dometto: I think this sadly doesn't work because `galaxy_gravity_command` is a default within the `galaxy` role, so it's not available in the playbook. But `{{ galaxy_venv_dir }}/bin/galaxyctl` might be a bit better?

hexylena: Ah I knew I was using that based on seeing it from somewhere: the built in handlers We have two handlers defined, that call this already (`handlers/gravity_23.0.yml`)

So these handlers as defined in the example are no longer necessary.

dometto: Ah right! Those handlers weren't working for me, but I now see that may be due to https://github.com/galaxyproject/ansible-galaxy/issues/202 -- so probably that issue should be fixed, and the handlers should not be defined in the Readme. :)

PR #212, opened 2024-04-16
Add GRAVITY_CONFIG_FILE to systemd env

dometto:
I noticed the reason why the molecule runs for version `23.0` were failing was that `supervisord` was attempting to run Galaxy with the wrong config file. On the failing containers, `/srv/galaxy/var/gravity/supervisor/supervisord.conf.d/_default_.d/galaxy_gunicorn_gunicorn.conf` has:

```
command         = /srv/galaxy/venv/bin/galaxyctl --config-file /srv/galaxy/server/config/galaxy.yml.sample exec _default_ gunicorn
```

It looks like this is because the systemd unit file doesn't correctly set the `GRAVITY_CONFIG_FILE` environment variable.

Additionally, I locally ran into some issues requiring further `apt` dependencies to be installed on Ubuntu in order for some of Galaxy's dependencies to compile correctly. This does not seem to have lead to test failures on CI, my hunch is that this due to me using an `ARM` based machine locally, but I'm not sure. I've added the extra `apt` dependencies that I needed as they seem pretty sane, but feel free to chuck out that commit, of course :)

### DIFF
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -19,6 +19,7 @@
     galaxy_user: galaxy
     galaxy_group: galaxy
     galaxy_privsep_user: gxpriv
+    galaxy_clone_depth: 1
     galaxy_config:
       galaxy:
         database_connection: sqlite:///{{ galaxy_mutable_data_dir }}/universe.sqlite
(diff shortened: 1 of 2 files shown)


dometto:
Can confirm that this passes the tests when running with `GALAXY_VERSION="23.0" FROM_IMAGE="ubuntu:22.04"` locally 👍

hexylena:
More tests passing is a fantastic improvement

Issue #213, opened 2024-05-02
Static files are not updated when using `galaxy_manage_subdomain_static`


mira-miracoli:
After we upgraded EU to 24.0, we noticed, that some subdomains were having old CSS files.
see (not public):
- https://github.com/usegalaxy-eu/issues/issues/537

This comes potentially from the [`ignore_existing`](https://github.com/galaxyproject/ansible-galaxy/blob/cc413b6e06d7695a90e51db5342f57931ddfa6b3/tasks/static_subdomain_dirs.yml#L47) flag I set to rsync the files from Galaxy's default static directory to the subdomain's static directory. The reason was, that ansible would touch files that are different for the subdomain on every plabook run. It would first overwrite them with the rsync command and than overwrite them again when it [copies over subdomain's custom file](https://github.com/galaxyproject/ansible-galaxy/blob/main/tasks/copy_static_files.yml).
My idea to solve this would be using this:
`{{ __galaxy_client_build_version }} != {{ __galaxy_current_commit_id }}`
as indicator that Galaxy has been updated and in that case omit the `ignore_existing` flag. Does that sound sensible to you
@hexylena @natefoo or do you think there is a better approach?

mira-miracoli:
@natefoo @hexylena any Ideas? maybe we can discuss it at GCC :)

hexylena:
sounds good, let's discuss then since I'm quite busy until then :)

PR #215, opened 2024-06-29
Only omit rsync, if Galaxy is not being updated

mira-miracoli:
should close 
- #213 
- https://github.com/usegalaxy-eu/issues/issues/500

(I can not test this currently, because my test instance is not working, so I would ask you to double check the syntax, please 🙏🏼 )


mira-miracoli:
(Tests are failing due to dependency issues, I think)

natefoo (approved):

PR #217, opened 2024-06-29
Custom subdomain static path

mira-miracoli:
requested by @natefoo 
for people who want to have the subdomain static dirs in separate locations (not Galaxy root)

(Again, not tested, because my test instance is broken)


mira-miracoli:
closed in favor of #216

PR #219, opened 2024-08-19
Make themes structured data instead of read from file

natefoo:
This way the role does not modify files in the playbook, and moves in line with other recent changes to move more of Galaxy's extraneous configs into structured data.

@mira-miracoli if you approve I'll update the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,60 +95,63 @@ Options below that control individual file or subdirectory placement can still o
 [custom]: vars/layout-custom.yml
 [fhs]: http://www.pathname.com/fhs/
 
-**New options for Galaxy 22.01 and later**
+**Process control with Gravity**
 
-The role can now manage the Galaxy service using [gravity][gravity]. This is the default for Galaxy 22.05 and later.
+The role can manage the Galaxy service using [gravity][gravity]. This is the default for Galaxy 22.05 and later.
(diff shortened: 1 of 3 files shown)


mira-miracoli:
cool, less files!
But the directory structure for the subdomain static files has to stay I guess?

mira-miracoli (approved):

natefoo:
In working on the documentation and defaults changes I noticed, I don't think we need the "default base domain" entry on the `*_base_host` config vars. For example, if you have:

```yaml
themes_config_file: themes_conf.yml
themes_config_file_by_host:
  usegalaxy.eu: themes_conf.yml
  assembly.usegalaxy.eu: assembly.yml
```

Then the `usegalaxy.eu` entry is redundant since, if there is no matching hostname for usegalaxy.eu found in `themes_config_file_by_host`, it would fall back to the value of `themes_config_file`. So I removed this bit, but if I have this wrong and it doesn't work like I think it does, let me know.

mira-miracoli:
I think it did not work without, but we can try :)

natefoo:
You're right, I have some additional hostnames set up for usegalaxy.org to reach my individual VMs directly (e.g. https://galaxy-web-05.galaxyproject.org/) and the themes aren't available via that route.

PR #221, opened 2024-11-12
fix welcome stat task

mira-miracoli:
that should have been obvious.. 🙄


bgruening:
Thanks @mira-miracoli!

PR #222, opened 2025-01-16
Add flexibility to galaxy_dirs, galaxy_privsep_dirs defaults

cat-bro:
My use case is that Galaxy Australia uses different ownership/permissions for the `galaxy_local_tools_dir` folder. With this change I'd be able to exclude that folder from `galaxy_privsep_dirs` without losing the valuable default lists.


jdavcs (approved):
This looks good to me. @cat-bro sorry for the delay - this fell through the cracks. Is this still needed?

cat-bro:
Thanks @jdavcs , yes, we still do have a use case for this.

PR #227, opened 2025-06-06
Don't build source maps

mvdbeek:


natefoo (approved):

martenson:
What is the workaround for when you want to debug JS on a deployed instance? Are the source maps going to be hosted elsewhere and linked somehow?

dannon:
Thanks!  We can also look at having the sourcemaps hosted separately (on the galaxy server) instead of inline as well -- just pulling them out dramatically trimmed down the bundles but I think it'd be fine to continue to host them as long as they're not a part of the initial load.

PR #232, opened 2025-09-01
Template subdomain static dirs using an Ansible module

kysrpex:
This Ansible role can manage subdomain static directories. The feature (introduced in PR #194) is enabled when `galaxy_manage_subdomain_static` is set to `true`.

The motivation of this PR is to improve the CI run times for usegalaxy.eu. At the moment, we are using the feature introduced in PR #198 that enables using symlinks to template subdomain static directories from Galaxy's static files directory (files specific to each subdomain are then merged on top of the template). It saves us a lot of disk space (since about ~154 MB are needed for each subdomain in a basic Galaxy setup from the GTN), but it two significant shortcomings: it generates very extensive logs (since each symlink needed to be checked and is logged by Ansible) and slows down playbook runs very noticeably (again, because each symlink is created by a separate Ansible task).

This commit moves the symlinking job to an Ansible module symlink_clone.py, so that only one task per subdomain is needed. Since that solves all shortcomings of the previous approach, not using symlinks is no longer an option.


defaults/main.yml
```diff
@@ -527,17 +527,7 @@ galaxy_themes_subdomains: []
...
 #             rgb(203, 119, 79) 100%)
-galaxy_themes_static_keys:
-  static_dir: ""
-  static_images_dir: "images/"
-  static_scripts_dir: "scripts/"
-  static_welcome_html: "welcome.html/"
-  static_favicon_dir: "favicon.ico"
-  static_robots_txt: "robots.txt"
 galaxy_themes_static_path: "{{ galaxy_root }}/server"
-galaxy_themes_static_dir: "{{ galaxy_root }}/server/static"
```
kysrpex, line 538: This was a "dead variable" (not used anywhere).

kysrpex:
@mira-miracoli Could you have a look at this? I see you have permission to merge to this repository.

jdavcs (approved):
Looks good to me.

Issue #240, opened 2026-01-09
Comments are getting dumped in optional requirement strings


microfuge:
Hi,
I am attempting a test install of galaxy (v25.1) and the ansible galaxy role playbook errors at the additional requirement strings. The comments added next to additional requirement strings are getting outputted which results in error with pip install. E.g This  `'fs.webdavfs>=0.4.2  # type: webdav'`

Steps  

* in galaxy virtualenv 
* `export PYTHONPATH=/path/to/galaxy/lib`
* in python console
```
import galaxy.dependencies
deps = galaxy.dependencies.optional("/path/to/galaxy.yml")
deps
['psycopg2-binary==2.9.10', 'ldap3==2.9.1', 'fs.webdavfs>=0.4.2  # type: webdav', 'webdavclient3 @ git+https://github.com/ezhov-evgeny/webdav-client-python-3@98c23d1abd15efc3db9cfc756429f00041578bc2']
```
* The pip fails with attempt `pip install 'fs.webdavfs>=0.4.2  # type: webdav'`

Just in case, error from ansible

<img width="1893" height="218" alt="Image" src="https://github.com/user-attachments/assets/81f13425-2555-4fb8-ba54-43ae92f25e0d" />

Thanks!

nsoranzo:
What Galaxy does in `scripts/common_startup.sh` is:

```shell
echo 'fs.webdavfs>=0.4.2  # type: webdav' | pip install -r /dev/stdin
```

which works fine in my testing (both with `pip` and `uv pip`). So this seems to me either a problem in the Ansible galaxy role, or something else.

microfuge:
Ah Thanks, 

Seems like the role directly installs via pip  https://github.com/galaxyproject/ansible-galaxy/blob/main/tasks/dependencies.yml and does not use echo.

microfuge:
Thanks again for taking care of this, I was also confused about which repo this issue belongs to. I still think that the best fix would be to have galaxy specify clean and standard requirement strings instead of using shell and whitespace separation. I also tend to single quote the requirement strings while doing pip installs.

mvdbeek:
Fixed in https://github.com/galaxyproject/galaxy/pull/21720

PR #243, opened 2026-01-29
Clean up Galaxy conditional dependencies before installation

tchaussepiedifb:
Removes inline `# type:` comments from conditional-requirements.txt so pip can install dependencies.   
This fixes the same error we encountered on `usegalaxy.fr` after switching to Docker runners with `alpine/ansible:2.16.1` for galaxy_25.0.

Refs: 
- https://github.com/galaxyproject/ansible-galaxy/issues/240
- https://github.com/galaxyproject/galaxy/pull/21593

### DIFF
--- a/tasks/dependencies.yml
+++ b/tasks/dependencies.yml
@@ -25,16 +25,20 @@
       register: conditional_dependencies
       changed_when: no
 
+    - name: Clean Galaxy conditional dependencies (remove inline comments)
+      set_fact:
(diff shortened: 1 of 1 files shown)


mvdbeek:
Thanks @tchaussepied, we've fixed this on the galaxy side by stripping the comments before printing the conditional dependencies (https://github.com/galaxyproject/galaxy/pull/21720).

PR #244, opened 2026-03-11
Fix bug templating subdomain static dirs via Ansible module

domgz:
This PR fixes two problems.

---

**`FileNotFoundError` when a directory doesn't exist in the target subdomain static dir**

If the target path does not exist, `dst_stat = dst_path.lstat()` leads to an error `FileNotFoundError: [Errno 2] No such file or directory: ...`.

Rearrange the line `dst_stat = dst_path.lstat()` and add a check based on `dst_path.exists()` for the case where `src_path.is_dir() == False` and `dst_path.exists() == False`; for example when the source is a file and the target does not exist.

Given that we are not dealing with file types other than symlinks, directories and regular files, that should be enough.

---

**Directory comparisons are not recursive.**

`filecmp.dircmp()` does not compare directories recursively. To detect if subdomain static dirs contents have changed, use a new function `compare_dirs()` that runs `filecmp.dircmp()` recursively.

### DIFF
--- a/library/symlink_clone.py
+++ b/library/symlink_clone.py
@@ -148,6 +148,26 @@ def set_permissions(
                 )
 
 
+def compare_dirs(src: Path, dst: Path) -> bool:
+    """Compare two directories recursively."""
+    comparison = filecmp.dircmp(src, dst)
+
+    changed = bool(
+        comparison.left_only
+        # or comparison.right_only  # target may contain extra files
+        or comparison.diff_files
+        or comparison.funny_files
+    )
+    if changed:
+        return True
+
+    for subdir in comparison.common_dirs:
+        if compare_dirs(src / subdir, dst / subdir):
+            return True
+
+    return False
+
+
 def run_module():
     """Run the Ansible module."""
     module = AnsibleModule(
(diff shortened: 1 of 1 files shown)


domgz:
@jdavcs this is a follow-up for #232.

The first error was detected because it was showing up on the CI, but I noticed also the second while fixing the former. It's surprising that it went under the radar for so long, especially considering it severly breaks the functionality (directories are only merged if a change is detected).

jdavcs (approved):
Great catch!
Works for me. I have just one request. It was not obvious to me what `compare_dirs` returns. I expected the opposite: `True` if there is no difference - i.e., similar to how `filecmp.cmp()` works. I don't switching to logic makes sense here since it's not a straightforward comparison like the `cmp()` function. However, maybe change the function name to something more explicit? (like `dirs_have_changed`?) Or, maybe even better, specify in the function docstring what it returns. Either one is fine - as long as it's obvious what it returns.

library/symlink_clone.py
```diff
@@ -144,6 +148,26 @@ def set_permissions(
                 )
 
 
+def compare_dirs(src: Path, dst: Path) -> bool:
+    """Compare two directories recursively."""
```
domgz:
```suggestion
    """Compare two directories recursively.
    
    Returns:
        True if the directories are different, False if they are considered
        identical.
    """
```

domgz:
Thanks for creating [the release](https://github.com/galaxyproject/ansible-galaxy/releases/tag/0.12.4)!

jdavcs:
@kysrpex new release imported to ansible galaxy. (I saw you updated your infrastructure playbook)

PR #246, opened 2026-03-23
Add a variable to restart handler command so that ‘graceful’ can be overridden

cat-bro:
The restart galaxy handler command being “galaxyctl graceful” is handy and it’s what we want most of the time, but there are occasions when I want to “restart” instead. For example, when Galaxy is switched off for maintenance, there is no benefit to the graceful restart and I’d like to just ‘galaxyctl restart’ by overriding `galaxy_gravity_restart_action`.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -418,6 +418,7 @@ __galaxy_gravity_pm: "{{ (galaxy_config_merged.gravity | default({})).process_ma
 __galaxy_gravity_instance_name: "{{ (galaxy_config_merged.gravity | default({})).instance_name | default(none) }}"
 galaxy_gravity_wrapper_path: "/usr/local/bin/galaxyctl{{ __galaxy_gravity_instance_name | ternary('-' ~ __galaxy_gravity_instance_name, '') }}"
 galaxy_gravity_command: "{{ galaxy_venv_dir }}/bin/galaxyctl"
+galaxy_gravity_restart_action: graceful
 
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -418,6 +418,7 @@ __galaxy_gravity_pm: "{{ (galaxy_config_merged.gravity | default({})).process_ma
 __galaxy_gravity_instance_name: "{{ (galaxy_config_merged.gravity | default({})).instance_name | default(none) }}"
 galaxy_gravity_wrapper_path: "/usr/local/bin/galaxyctl{{ __galaxy_gravity_instance_name | ternary('-' ~ __galaxy_gravity_instance_name, '') }}"
 galaxy_gravity_command: "{{ galaxy_venv_dir }}/bin/galaxyctl"
+galaxy_gravity_restart_action: graceful
```
mira-miracoli: maybe we can add which options are allowed?
```suggestion
# Gravity restart behaviour, one of: restart, graceful
# Documentation: https://gravity.readthedocs.io/en/stable/basic_usage.html#managing-a-production-galaxy
galaxy_gravity_restart_action: graceful 
```

cat-bro: lovely, thanks @mira-miracoli !

mira-miracoli (approved):
Looks good to me :)

natefoo (approved):

PR #248, opened 2026-04-17
Fix Galaxy version file path after upstream reorganization

### LINKED ISSUE #247: Galaxy `lib/galaxy/version.py` was renamed
to `lib/galaxy/version/__init__.py` in https://github.com/galaxyproject/galaxy/pull/21977

This breaks the "Collect Galaxy version file" task in `tasks/_inc_galaxy_version.yml` .

ksuderman:
This PR fixes a breaking change introduced when Galaxy reorganized their version module structure. Galaxy moved `lib/galaxy/version.py` to `lib/galaxy/version/__init__.py` in their codebase, which broke the "Collect Galaxy version file" task in this Ansible role.

The fix implements backward compatibility by checking for the new file location first (`lib/galaxy/version/__init__.py`) and falling back to the legacy location (`lib/galaxy/version.py`) if the new one doesn't exist. This ensures the role works seamlessly with both older and newer versions of Galaxy. Proper error handling is also added to provide a clear message if neither file location is found.

This change affects multiple tasks that depend on Galaxy version detection, including mutable setup, static setup, systemd configuration, and client build processes.

Closes #247

### DIFF
--- a/tasks/_inc_galaxy_version.yml
+++ b/tasks/_inc_galaxy_version.yml
@@ -14,7 +14,7 @@
   register: __galaxy_version_legacy_stat
   when: not __galaxy_version_init_stat.stat.exists
 
-- name: Collect Galaxy version file (new location)
+- name: Collect Galaxy version file for Galaxy 26.1+
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
   register: __galaxy_version_file
--- a/tasks/_inc_galaxy_version.yml
+++ b/tasks/_inc_galaxy_version.yml
@@ -3,33 +3,22 @@
 
 # Currently only used by mutable config setup but placed in an include because it'll probably be used by more
 
-- name: Check version file location for Galaxy 26.1+ 
-  stat:
-    path: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
-  register: __galaxy_version_init_stat
-
-- name: Check for legacy Galaxy version file location
-  stat:
-    path: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
-  register: __galaxy_version_legacy_stat
-  when: not __galaxy_version_init_stat.stat.exists
-
-- name: Collect Galaxy version file for Galaxy 26.1+
+- name: Collect Galaxy version file (Galaxy 26.1+)
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
   register: __galaxy_version_file
-  when: __galaxy_version_init_stat.stat.exists
+  failed_when: false
 
 - name: Collect Galaxy version file (legacy location)
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
   register: __galaxy_version_file
-  when: not __galaxy_version_init_stat.stat.exists and __galaxy_version_legacy_stat.stat.exists
+  when: __galaxy_version_file is not defined or __galaxy_version_file.failed
 
 - name: Fail if Galaxy version file not found
   fail:
-    msg: "Galaxy version file not found at {{ galaxy_server_dir }}/lib/galaxy/version/__init__.py or {{ galaxy_server_dir }}/lib/galaxy/version.py"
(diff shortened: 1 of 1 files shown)


tasks/_inc_galaxy_version.yml
```diff
@@ -3,10 +3,33 @@
 
 # Currently only used by mutable config setup but placed in an include because it'll probably be used by more
 
-- name: Collect Galaxy version file
+- name: Check for new Galaxy version file location
```
afgane:
```suggestion
- name: Check version file location for Galaxy 26.1+ 
```

tasks/_inc_galaxy_version.yml
```diff
@@ -3,10 +3,33 @@
...
+    path: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
+  register: __galaxy_version_init_stat
+
+- name: Check for legacy Galaxy version file location
+  stat:
+    path: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
+  register: __galaxy_version_legacy_stat
+  when: not __galaxy_version_init_stat.stat.exists
+
+- name: Collect Galaxy version file (new location)
```
afgane:
```suggestion
- name: Collect Galaxy version file for Galaxy 26.1+
```

afgane (approved):

tasks/_inc_galaxy_version.yml
```diff
@@ -3,10 +3,33 @@
...
+- name: Collect Galaxy version file (legacy location)
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
   register: __galaxy_version_file
+  when: not __galaxy_version_init_stat.stat.exists and __galaxy_version_legacy_stat.stat.exists
+
+- name: Fail if Galaxy version file not found
+  fail:
+    msg: "Galaxy version file not found at {{ galaxy_server_dir }}/lib/galaxy/version/__init__.py or {{ galaxy_server_dir }}/lib/galaxy/version.py"
+  when: not __galaxy_version_init_stat.stat.exists and not __galaxy_version_legacy_stat.stat.exists
```
domgz:
```suggestion
- name: Check version file location for Galaxy 26.1+ 
  stat:
    path: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
  register: __galaxy_version_init_stat

- name: Collect Galaxy version file for Galaxy 26.1+
  slurp:
    src: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
  register: __galaxy_version_file
  when: __galaxy_version_init_stat.stat.exists

- name: Collect Galaxy version file (legacy location)
  slurp:
    src: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
  register: __galaxy_version_file
  when: not __galaxy_version_init_stat.stat.exists
```

I'm only suggesting the change because in the past we were fine with assuming that the file exists, so probably we are still fine with it.

ksuderman: Well, that assumption being wrong is what led me here, so I think explicit error handling is not a bad thing.  Having said that, and looking at your suggested changes, I think this bit can be tidied up somewhat.

tasks/_inc_galaxy_version.yml
```diff
@@ -3,10 +3,22 @@
 
 # Currently only used by mutable config setup but placed in an include because it'll probably be used by more
 
-- name: Collect Galaxy version file
+- name: Collect Galaxy version file (Galaxy 26.1+)
+  slurp:
+    src: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
+  register: __galaxy_version_file
+  failed_when: false
```
domgz, line 10: @mvdbeek I was reviewing the PR, it is broken.

That line makes `__galaxy_version_file.failed` always false.

```
TASK [Debug] *****************************************************************************************************************************************************************
ok: [sn09.galaxyproject.eu] => {
    "__galaxy_version_file": {
        "changed": false,
        "failed": false,
        "failed_when_result": false,
        "msg": "file not found: /opt/galaxy/server/lib/galaxy/version/__init__.py"
    }
}
```

domgz: Please see https://github.com/galaxyproject/ansible-galaxy/pull/249.

PR #250, opened 2026-04-21
Use ignore_errors rather that failed_when

### LINKED ISSUE #247: Galaxy `lib/galaxy/version.py` was renamed

ksuderman:
A corrected fix for the breaking change introduced when Galaxy reorganized their version module structure.

The previous fix (PR #248) was broken because it used `failed_when: false`, which caused the slurp task to "succeed" even when the file didn't exist, but without populating the `content` field. This led to failures in the subsequent "Determine Galaxy version" task.

This corrected implementation uses `ignore_errors: true` instead, which allows the task to fail naturally when the file doesn't exist, but continues playbook execution. This ensures that when a file is successfully read, the `content` field is properly populated, and when it fails, the `failed` field is correctly set to `true`.

The fix tries the new file location first (`lib/galaxy/version/__init__.py`) and falls back to the legacy location (`lib/galaxy/version.py`) if the first attempt fails. This ensures backward compatibility with both older and newer versions of Galaxy.

Closes #247

### DIFF
--- a/tasks/_inc_galaxy_version.yml
+++ b/tasks/_inc_galaxy_version.yml
@@ -7,18 +7,13 @@
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version/__init__.py"
   register: __galaxy_version_file
-  failed_when: false
+  ignore_errors: true
 
 - name: Collect Galaxy version file (legacy location)
   slurp:
     src: "{{ galaxy_server_dir }}/lib/galaxy/version.py"
   register: __galaxy_version_file
-  when: __galaxy_version_file is not defined or __galaxy_version_file.failed
-
-- name: Fail if Galaxy version file not found
-  fail:
-    msg: "Galaxy version file not found in expected locations"
-  when: __galaxy_version_file.failed is defined and __galaxy_version_file.failed
+  when: __galaxy_version_file.failed
 
 - name: Determine Galaxy version
   set_fact:


domgz (commented):
Thanks, this works as expected.

```
PLAY [UseGalaxy.eu] ************************************************************

TASK [Gathering Facts] *********************************************************
ok: [sn09.galaxyproject.eu]

TASK [Collect Galaxy version file (Galaxy 26.1+)] ******************************
fatal: [sn09.galaxyproject.eu]: FAILED! => {"changed": false, "msg": "file not found: /opt/galaxy/server/lib/galaxy/version/__init__.py"}                   
...ignoring

TASK [Collect Galaxy version file (legacy location)] ***************************
ok: [sn09.galaxyproject.eu]

TASK [Determine Galaxy version] ************************************************
ok: [sn09.galaxyproject.eu]

TASK [Debug] *******************************************************************
ok: [sn09.galaxyproject.eu] => {
    "__galaxy_major_version": "26.0"
}

PLAY RECAP *********************************************************************
sn09.galaxyproject.eu      : ok=5    
```

domgz (approved):

ksuderman:
@mvdbeek

PR #252, opened 2026-04-24
Docker tests

mvdbeek:


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,75 @@
+---
+name: Docker Build Tests
+
+on:
+  pull_request:
+  push:
```
nsoranzo, line 6: Would it make sense to also schedule this once a week to catch Galaxy breaking changes early, like we do for the BioBlend tests ( https://github.com/galaxyproject/bioblend/blob/main/.github/workflows/test.yaml#L5-L7 )?

dannon (approved):

PR #256, opened 2026-07-30
Update iframe URL to remove file extension

bgruening:


tasks/static_subdomain_dirs.yml
```diff
@@ -24,7 +24,7 @@
 
 - name: Check if iframe for {{ subdomain.name }} exists  # noqa name[template]
   ansible.builtin.uri:
-    url: "{{ galaxy_themes_welcome_url_prefix }}{{ subdomain.name }}.html"
+    url: "{{ galaxy_themes_welcome_url_prefix }}{{ subdomain.name }}"
```
bgruening, line 27:
```suggestion
    url: "{{ galaxy_themes_welcome_url_prefix }}{{ subdomain.name }}/index.html"
```

Or this one. 
Both are needed for the new sub-domains served from the Hub. E.g. https://galaxyproject.org/bare/eu/usegalaxy/earth-system/

domgz: Suggestion to keep this role "agnostic" for any kind of deployment: define `galaxy_themes_welcome_url_suffix` or similar.

bgruening: Good idea, just not sure of anyone else is using the subdomains in this way.

If I don't get to it today, can you maybe work on it next week?

